PR #1, opened 2018-06-24
Added Changes for release

matsduf:


sandoche2k:
Maybe we should add that there is no perl interface when compared to the old GUI and also we need to install apache.

matsduf:
@sandoche2k, please review again after update.

PR #7, opened 2018-09-11
Update result CSS: set a cursor pointer on expandable results

chtitux:
The expandable lines can be expanded by clicking on them.
Set a "pointer" cursor (aka "click cursor") for such lines, so user is aware he can click on them.


sandoche2k:
@pamasse do you agree ?

PR #10, opened 2018-12-14
Get dynamically profiles from the backend

pamasse:
Get dynamically profiles from the backend through the new API function 'profiles_names'.


sandoche2k:
@mattias-p @matsduf @mtoma check and approve

PR #12, opened 2018-12-14
Fix #8 - Change HTML DS types (keytag and digest) from String to Int as expected in the API

### LINKED ISSUE #8: Pre-delegated DNSSEC check: DS: "expected integer, got string"
When filling in the form to do a pre-delegation check with DNSSEC, an `Internal Server Error` message pops up. It seems like the API is giving the following error message:

```json
{
   "jsonrpc":"2.0",
   "error":{
      "data":"/ds_info/0/algorithm: Expected integer - got string. /ds_info/0/digtype: Expected integer - got string. /ds_info/0/keytag: Expected integer - got string. /ds_info/0/algorithm: Expected integer - got string. /ds_info/1/digtype: Expected integer - got string. /ds_info/0/keytag: Expected integer - got string. /ds_info/0/algorithm: Expected integer - got string. /ds_info/2/digest: String does not match '^[A-Fa-f0-9]{40}$|^[A-Fa-f0-9]{64}$' /ds_info/2/digtype: Expected integer - got string. /ds_info/0/keytag: Expected integer - got string.\n",
      "message":"Invalid method parameter(s).",
      "code":"-32602"
   },
   "id":null
}
```
The data sent for `keytag` and `digtype` was an integer though. So it seems like the API is checking the data typesafe and should try to convert the strings "56482" and "1" to integers.

pamasse:
Fix #8 - Change HTML DS types (keytag and digest) from String to Int as expected in the API.
Replace html value for select options to angular [value] which cast string to int.

### DIFF
--- a/src/app/components/form/form.component.html
+++ b/src/app/components/form/form.component.html
@@ -109,21 +109,21 @@
               <div class="nsForm" *ngFor="let itemrow of digestForm.controls.itemRows.controls; let i=index" [formGroupName]="i">
                 <input formControlName="keytag" class="form-control" name="form.keytag" placeholder="Key tag">
                 <select formControlName="algorithm" class="form-control" name="form.algorithm">
-                  <option value="3">DSA/SHA1</option>
-                  <option value="5">RSA/SHA1</option>
-                  <option value="6">DSA-NSEC3-SHA1</option>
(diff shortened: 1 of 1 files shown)


sandoche2k:
@mattias-p @matsduf @mtoma check and approve

sandoche2k:
@pamasse pls. merge

PR #19, opened 2019-01-27
Merge master into develop

matsduf:
There are commits in the master branch that need to be merge into develop to make develop be a superset.


pamasse:
@sandoche2k It could be a good think to create a travis test or something like this for the GUI too.

matsduf:
@sandoche2k, yes it is possible to get the zip file. If you select the develop branch in the repository and then click on "clone or download" and select zip file and you get the same thing, but based on the HEAD commit of the develop branch.

Issue #26, opened 2019-01-31
Define supported environments


matsduf:
For GUI we need to define what browsers (and version of that) on what OS that we support or at least test on.

pamasse:
Hi @matsduf,

I've done some test with different tools like [browserstack](www.browserstack.com). With this kind of tool we can automate the test of the gui very easily (because the gui is very simple).

I ask for a free plan (https://www.browserstack.com/open-source) for zonemaster GUI. But, to accept the free plan, they would like their logo on the zonemaster gui readme page. 

What do you think about this?

sandoche2k:
Could be closed when the PR https://github.com/zonemaster/zonemaster/pull/756 is closed? @pamasse

matsduf:
Defined in zonemaster/zonemaster.

PR #37, opened 2019-04-02
Add some e2e test with selenium (only works on local)

pamasse:
Not a full response to #26, but this is a solution to select a list of supported browser.

Here some selenium test with a pre implementation of the BrowserStack plateform link to run parallel tests.


matsduf:
I trust that @pamasse has done this correctly.

@pamasse, what does "only works on local" refer to?

PR #41, opened 2019-04-29
Add a tests before fetching information about parent zone

### LINKED ISSUE #39: Tries to fetch from parent zone even if domain has not been given
If you are in the pre-delegated domain mode and leave the domain name field empty and press "fetch data from parent zone" the gui will request the backend for "get_data_from_parent_zone" with empty domain. 

Maybe it relates to issue #38.

pamasse:
Check before fetching information about parent zone if the domain input is not empty

Fix #39


src/app/components/form/form.component.ts
```diff
@@ -121,6 +117,16 @@ export class FormComponent implements OnInit {
     return selectedItems.length ? selectedItems : null;
   }
 
+  private displayDataFromParent() {
+
+    if (this.form['domain'] === '') {
+      this.alertService.error('Domain name required');
```
matsduf: A problem with this solution is that we always get the message in English, don't we? How can we have it translatable to the supported languages?

pamasse: The system is already in place. We need to put the translation in the lang files (src/assets/i18n/[fr/en/sv]) with the english text as the json key. I started this work for french but i'm not very confident with my Swedish :).
I create an issue for that #44

matsduf: What do you mean by "the system is already in place"? This is new code that has not yet been merged to develop branch. We are still discussing it.

pamasse: The code bellow send a string (here 'Domain name required') to the error method of the alert service. The alert service display the text with the proper settings (color red for error, orange for warning etc.) and also try to translate the text. If the translation is not define, it keep the string (here 'Domain name required') and display it. To see this system in action, switch to french and try to run a test without domain name. 

Yes, i will add the English translation in a separate PR.

PR #47, opened 2019-05-09
Fix sending empty ns when the advanced options are not open

pamasse:
Fix sending empty ns when the advanced options are not open


pamasse:
The commit fix the issue

PR #53, opened 2019-05-17
Improved cut'n'paste friendliness for release testing

mattias-p:
Release testing is more streamlined if the `unzip` refers to the same file name as the file created by the `npm run release` command.

PR #54, opened 2019-05-17
Updated installation instructions on FreeBSD:

matsduf:
* Corrected path for included configuration files.
* Added needed enabling of modules in Apache.
* Removed uneeded use of wget for fetch instead (and removed
  installation of wget)
* Editorial updates

The current instructions are incomplete or wrong.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -14,6 +14,10 @@ Before installing Zonemaster Web GUI, you should [install Zonemaster::Engine][
 Zonemaster::Engine installation] and [Zonemaster::Backend][Zonemaster::Backend
 installation].
 
+Prerequisite for FreeBSD is that the package system is upadated and activated,
+see FreeBSD section of [install Zonemaster::Engine].
+
+
 ## Installation
 
 This instruction covers the following operating systems:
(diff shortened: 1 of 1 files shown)


docs/Installation.md
```diff
@@ -109,50 +113,47 @@ sudo systemctl reload apache2
...
-```
+* Start Apache:
+ 
+``service apache24 start``
 
 * Install Zonemaster Web GUI
 ```sh
-wget https://github.com/zonemaster/zonemaster-gui/releases/download/v3.0.1/zonemaster_web_gui.zip -O temp.zip
-mkdir -p  /var/www/html/zonemaster-web-gui
+fetch https://github.com/zonemaster/zonemaster-gui/releases/download/v3.0.1/zonemaster_web_gui.zip
```
sandoche2k: Should be 3.0.2

sandoche2k:
@pamasse since the file name now is not "temp.zip", you need to change for all distributions as @matsduf has done

docs/Installation.md
matsduf: That will be fixed at release time. We should probably have 3.0.3 since we have added things after.

matsduf:
@sandoche2k, that is done in #53.

matsduf:
Rebased after #53 was merged to develop branch.

Issue #48, opened 2019-05-09
Remove 14.0.4 Ubuntu requirements from install instructions


sandoche2k:

sandoche2k:
@pamasse can you update and close?

PR #62, opened 2019-05-20
Stop referring to unsupported OS version

mattias-p:
Ubuntu 14.04 has reached its end of life and we don't support it anymore.


pamasse:
This PR fix  #48

Issue #56, opened 2019-05-18
Warnings when creating zip file


matsduf:
The zip file is created on the develop branch on the following commit:
```
commit fc3be71d73cb8e0c56633ae88684bbe2e34a7184
Merge: 8717a69 9dec49a
Author: pamasse <pa.masse@outlook.fr>
Date:   Fri May 17 16:02:01 2019 +0200

    Merge pull request #47 from pamasse/fix-empty-ns-default-test
    
    Fix sending empty ns when the advanced options are not open
```

Running `npm install` creates warnings. Can they be ignored? Can warnings always be ignored?
```
$ npm install
npm WARN @angular/compiler-cli@5.2.5 requires a peer of typescript@>=2.4.2 <2.7 but none is installed. You must install peer dependencies yourself.
npm WARN ajv-keywords@3.1.0 requires a peer of ajv@^6.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN bootstrap@4.0.0 requires a peer of jquery@1.9.1 - 3 but none is installed. You must install peer dependencies yourself.
npm WARN bootstrap@4.0.0 requires a peer of popper.js@^1.12.9 but none is installed. You must install peer dependencies yourself.
npm WARN tsickle@0.26.0 requires a peer of typescript@>=2.4.2 <2.7 but none is installed. You must install peer dependencies yourself.
npm WARN zonemaster-gui@3.0.1 No repository field.
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.1.3 (node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.1.3: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})

updated 1 package and audited 10861 packages in 18.887s
found 293 vulnerabilities (29 low, 178 moderate, 84 high, 2 critical)
  run `npm audit fix` to fix them, or `npm audit` for details
```
Running `npm run release` also creates some warnings. Can they be ignored?
```
$ npm run release

> zonemaster-gui@3.0.1 release /home/matsd/git/zonemaster-gui
> npm run build && node scripts/create_release.js

npm WARN invalid config loglevel="notice"

> zonemaster-gui@3.0.1 build /home/matsd/git/zonemaster-gui
> node scripts/faq.js && ng build -prod

[...]

PR #76, opened 2019-10-08
Update dependencies (Angular and co)

pamasse:
Big update of dependencies (angular v5 to v8, etc.).
Manually tested to be sure to not have functional regression.

The build process do not produce warning message anymore as explain in #56


README.md
```diff
@@ -91,6 +91,11 @@ Basically, the GUI is a serverless Angular application which use the Backend JSO
 
 Other than the installation file, the [docs directory](docs/), contains a translation guide for the GUI
 
+#### Security of our dependencies
```
mattias-p: This is a really good clarification!
It should be checked at least for every release, right? Could you make a PR to update the ReleaseProcess document in the main repo?

matsduf:
Will resolve issue #56.

README.md
pamasse: Every time we update Perl!
But yes, ideally, at least for every release.

PR #111, opened 2020-04-06
Update dependencies

pamasse:
It's an update of dependencies.
The big change concerns our main framework, from Angular 8 to Angular 9.

To apply the changes, please delete the `node_modules` folder and run `npm install`

Fix #108 and #109

PR #112, opened 2020-04-07
update tsconfig for ivy engine

pamasse:
Update tsconfig.app.ts file in order to be compliant with the new angular core engine, ivy.

It removes the WARNING message:
`WARNING in zonemaster-gui/src/environments/environment.prod.ts is part of the TypeScript compilation but it's unused.
Add only entry points to the 'files' or 'include' properties in your tsconfig.`

### DIFF
--- a/src/tsconfig.app.json
+++ b/src/tsconfig.app.json
@@ -5,6 +5,11 @@
     "baseUrl": "./",
     "types": []
   },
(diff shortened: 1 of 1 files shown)


matsduf:
Addresses the issue documented in https://github.com/zonemaster/zonemaster-gui/pull/111#issuecomment-609894816.

PR #117, opened 2020-04-21
Added space after string

matsduf:
Added space where it should be to make output correct.


matsduf:
@pamasse, I just need your approval so that I can merge.

PR #120, opened 2020-04-23
New logotype for Zonemaster (GUI)

matsduf:
Moved to a new logotype. Removing old, adding new and updating references.

The file here is a copy of the file added to the [zonemaster/zonemaster] by PR zonemaster/zonemaster#844. And we keep the same name.

[zonemaster/zonemaster]: https://github.com/zonemaster/zonemaster


matsduf:
@sandoche2k, please note that this is a second PR that had the same name.

matsduf:
@pamasse, please also check if anything else needs to be updated. I searched for the current logo and replaced with the new where I found a reference to it.

PR #139, opened 2020-08-28
type profil

rsau:


matsduf:
@pamasse, this should be fine, shouldn't it?

pamasse:
> @pamasse, this should be fine, shouldn't it?

Hi mats, no this is a static typo correction, we should use the i18n module in order to translate the word in all language. 

If I check with google translate: 
fr: profil
en: profile
sv: profil
da: profil

pamasse:
There are an issue with the version of the chrome driver to test the application. This PR is correct and can be merged.

matsduf (approved):

pamasse:
@matsduf who do the merge?

PR #156, opened 2020-11-07
For release v2020.1 (GUI)

matsduf:
* Changes: Updated with changes since last release.
* Installation.md: Updated version string
* package.json, environment.prod.ts and environment.ts: Set new version in code

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,5 +1,14 @@
 Release history for Zonemaster component Zonemaster-GUI 
 
+v3.2.1 2020-11-06 (public release version)
+
+ [Fixes]
+ - Correct fragment in links to FAQ (#148, #147)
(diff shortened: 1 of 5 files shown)


matsduf:
There is also package-lock.json with version string, but it will be generated by `npm install`. Maybe it should then be pushed to the repository.

sandoche2k (approved):

pnax (commented):
About the `package-lock.json`, why not adding it to this PR ?

matsduf:
> About the `package-lock.json`, why not adding it to this PR ?

As I understand the reply from @pamasse the file is recreated when doing `npm install` in the build zip file process. And that will be done after this PR has been merged, or else the version number will be wrong. I suggest that we look at it after the release and possibly merge it in to the develop branch then.

Issue #161, opened 2020-12-16
Make i18n/*.json consistent


matsduf:
This issue replaces the second half of #143.

In most cases the key from the code is the English text to be displayed, and that is straight forward, e.g. "Domain name required". When you translated you start from the key and produce the message in the target language. In English, key and value should be the same in [en.json] (except for next case).

In some cases the key is a key for a long text, e.g. "About DNS Text", and there is a longer text in all xx.json files, also in the en.json. The key, then, should be more code like, e.g. "ABOUT_DNS_TEXT" to make it more obvious that there is a "translation" even to English. In general, there should be no more problem with these cases.

The third category are cases where the key could be a message, but is "translated" to a different messages also in Enlish, e.g. "About ZoneMaster" --> "About Zonemaster". We should really get rid of those. In en.json there should only be the two first categories, key == value or key == SPECIAL_KEY.

Also see the discussion in https://github.com/zonemaster/zonemaster-gui/pull/121#issuecomment-699990680.

[en.json]: https://github.com/zonemaster/zonemaster-gui/blob/develop/src/assets/i18n/en.json

matsduf:
Resolved by #162.

PR #162, opened 2020-12-16
Resolve issue #161

matsduf:
Update string in code and matching string in translation files.

en.json should either have same key and value, or having a key like key.

Other *.json are updated to be consistent.

@pamasse, can you especially verify that `e2e/FR10.e2e-spec.ts` is still correct after my update.


sandoche2k (approved):

pnax (approved):

matsduf:
Resolves #161.

PR #166, opened 2020-12-16
Remove third party stylesheet

### LINKED ISSUE #151: Is the prism CSS stylesheet still needed ?
Currently there is one CSS stylesheet loaded from third party, see https://github.com/zonemaster/zonemaster-gui/blob/develop/src/index.html#L7

Commit https://github.com/zonemaster/zonemaster-gui/commit/0d0ddc48edf40ee545715b9b7b7a8d90e8234015 removes Prism dependencies. And it seems that the code is free from any references to the rules added by this stylesheet.

Therefore, is this stylesheet still needed ? (it would be nice to avoid calls to third parties)

pnax:
It looks like the Prism stylesheet is not used anywhere. So it seems safe to remove it.
Fixes #151

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -4,7 +4,6 @@
   <meta charset="utf-8">
   <title>Zonemaster</title>
 
-  <link href="https://cdnjs.cloudflare.com/ajax/libs/prism/1.6.0/themes/prism.min.css" rel="stylesheet" />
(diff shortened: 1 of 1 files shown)


matsduf (approved):

matsduf:
@pnax, please merge.

PR #165, opened 2020-12-16
Updating translation guide and adding Norwegian to GUI

matsduf:
The first commit updates the translation guide. The second commit adds Norwegian ("nb") by using the updated guide. The entries have also been sorted on language code.

PR #169, opened 2021-01-05
Check if browser language is supported

### LINKED ISSUE #168: Web browsers in other languages - no data for test results
When using web browsers set with other languages than the one supported by Zonemaster, the results are not displayed.
The following error message is displayed : `No data for this test` but the data does exist.

The query sent to the api is wrong as it embarks the browser language which may be unknown to the backend.

If the browser is using an unsupported language, the results should be displayed in English. In this situation the query should set the `language` parameter to `en`.

See https://github.com/zonemaster/zonemaster/issues/918

pnax:
Do not blindly use the browser language. Verify that the GUI supports it, or force English.

Addresses https://github.com/zonemaster/zonemaster/issues/918
Closes #168

### DIFF
--- a/src/app/components/navigation/navigation.component.ts
+++ b/src/app/components/navigation/navigation.component.ts
@@ -13,11 +13,16 @@ export class NavigationComponent implements OnInit {
   public isShrunk = false;
   public activeBackToTop = false;
   public lang = 'en';
+  private lang_default = 'en';
 
   constructor(private translateService: TranslateService, zone: NgZone) {
-    this.translateService.setDefaultLang(this.lang);
+    this.translateService.setDefaultLang(this.lang_default);
     this.lang = this.translateService.getBrowserLang();
-    this.translateService.use(this.lang);
+    if (this.isValidLanguage(this.lang)) {
+      this.setLanguage(this.lang);
+    } else {
+      this.setLanguage(this.lang_default);
+    }
     this.logoUrl = AppService.getLogoUrl();
     window.onscroll = () => {
       zone.run(() => {
@@ -44,5 +49,10 @@ export class NavigationComponent implements OnInit {
 
   public setLanguage(lang: string) {
     this.translateService.use(lang);
+  }
+
+  private isValidLanguage(lang: string) {
+    const validLanguages = [ 'en', 'fr', 'sv', 'da' ];
+    return validLanguages.includes(lang);
   }
 }
--- a/src/app/components/navigation/navigation.component.ts
+++ b/src/app/components/navigation/navigation.component.ts
@@ -52,7 +52,7 @@ export class NavigationComponent implements OnInit {
   }
 
   private isValidLanguage(lang: string) {
-    const validLanguages = [ 'en', 'fr', 'sv', 'da' ];
+    const validLanguages = [ 'da', 'en', 'fr', 'sv' ];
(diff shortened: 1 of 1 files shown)


src/app/components/navigation/navigation.component.ts
```diff
@@ -45,4 +50,9 @@ export class NavigationComponent implements OnInit {
   public setLanguage(lang: string) {
     this.translateService.use(lang);
   }
+
+  private isValidLanguage(lang: string) {
+    const validLanguages = [ 'en', 'fr', 'sv', 'da' ];
```
matsduf, line 55: In PR #165 (not yet merged) I update all places where the language tags are mentioned to be in alphabetical order to have it consistent. I suggest that they are listed as
```
    const validLanguages = [ 'da', 'en', 'fr', 'sv' ];
```

matsduf (commented):
For me this looks correct. I let @pamasse comment before I approve formally.

When this is merged I must remember to update #165.

src/app/components/navigation/navigation.component.ts
pnax: Fixed

pnax:
> When this is merged I must remember to update #165.

Depending on which one is merged first, this can also be done here.

And this also adds a default variable, which might need a change in the documentation in the [Change default language section](https://github.com/zonemaster/zonemaster-gui/blob/ac535b101bca0324e42b6af399ea18594545c116/docs/TranslationGuide.md#change-default-language)

matsduf (approved):

matsduf:
@pamasse, can you review this?

src/app/components/navigation/navigation.component.ts
```diff
@@ -13,11 +13,16 @@ export class NavigationComponent implements OnInit {
   public isShrunk = false;
   public activeBackToTop = false;
   public lang = 'en';
+  private lang_default = 'en';
```
matsduf, line 16: @pnax, if `private lang_default` should always be equal to `public lang` can't you just make them equal in the code? Or is there some use case for them being different?

pnax: Well, the current code logic need 2 distinct variables because `this.lang` will be overwritten by the current browser language at [line 20 (b35695e)](https://github.com/zonemaster/zonemaster-gui/blob/b35695e3d5c688052f3154ee953d7a3622765a57/src/app/components/navigation/navigation.component.ts#L20) and I needed to keep track of the default language somewhere to set it as a fallback [line 24 (b35695e)](https://github.com/zonemaster/zonemaster-gui/blob/b35695e3d5c688052f3154ee953d7a3622765a57/src/app/components/navigation/navigation.component.ts#L24).

Looking at [line 19 (b35695e)](https://github.com/zonemaster/zonemaster-gui/blob/b35695e3d5c688052f3154ee953d7a3622765a57/src/app/components/navigation/navigation.component.ts#L19) and the call to [`setDefaultLang`](https://github.com/ngx-translate/core#methods), we can think that all this logic is not needed. But somehow the default language is not used as a fallback if the browser language does not match any supported languages. This is incidentally the issue that was raised. To be honest I don't know much of Angular nor NGX-translate, so I coded the logic and kept the current code in case. Maybe the `setDefaultLang` call can be removed.

PR #172, opened 2021-01-25
Centos installation

mattias-p:
The instruction is updated to:
* Relieve users of from having to figure out that they need to disable/configure SELinux and how to do it.
* Enable remote access to the Web GUI by opening up the firewall.
* Clarify that editing zonemaster.conf is optional, and provide a cut-and-pasteable command.
* Give more feedback and require less interaction.
* Fix the headings.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -59,7 +59,14 @@ sudo setsebool -P httpd_can_network_connect=1
 sudo install -v /var/www/html/zonemaster-web-gui/zonemaster.conf-example /etc/httpd/conf.d/zonemaster.conf
 ```
 
-Then update the zonemaster.conf file with your own ServerName, ServerAlias, ServerAdmin
+Optionally update the zonemaster.conf.
+E.g. if Zonemaster-Backend RPCAPI runs on another server or on another port (not
+port 5000), update ProxyPass and ProxyPassReserve.
+Or if you want provide your own settings for ServerName, ServerAlias and ServerAdmin.
+
+```sh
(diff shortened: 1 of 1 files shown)


matsduf (dismissed):

docs/Installation.md
```diff
@@ -31,42 +31,54 @@ This instruction covers the following operating systems:
...
 ```sh
 sudo yum update
 sudo yum install httpd
 ```
 
+
 #### Install Zonemaster Web GUI
 
 ```sh
 wget https://github.com/zonemaster/zonemaster-gui/releases/download/v3.2.1/zonemaster_web_gui.zip -O zonemaster_web_gui.zip
```
pnax, line 45: Can you make the URL the last argument here ?

pnax (commented):
Looks good. Just a side suggestion on the argument order passed to `wget`.

docs/Installation.md
mattias-p: I suppose. Why is that better?

pnax: I would say that usually options come first on the command line, and this is how [the syntax is defined-(https://www.gnu.org/software/wget/manual/wget.html#Invoking). Moreover if you copy paste the command, you want to know where the file is stored more than where it comes from (but this is last argument might be completely subjective).

However, I see you updated the PR in a way that makes this remark irrelevant.

pnax (approved):

matsduf (approved):

Issue #181, opened 2021-02-12
Logotype does not display correctly on Chrome (and Chromium)


matsduf:
When Zonemaster-GUI is opened the Zonemaster logotype with built-in Afnic and Internetstiftelsen logotypes appear in top left. That works with Firefox and Safari, but it does not work with Chrome.

## Screenshot from Firefox:
![logo-is-displayed-by-Firefox](https://user-images.githubusercontent.com/5430733/107759551-c4095f00-6d28-11eb-8505-fc1eeea6909f.png)

## Screenshot from Chrome:
![logo-is-not-displayed-by-chrome](https://user-images.githubusercontent.com/5430733/107759600-d4b9d500-6d28-11eb-9d7d-86e1028e7175.png)

pamasse:
Try with a PNG image sized 250x63px

matsduf:
#187 resolves this issue.

pnax:
Just to inform you that I have made a quick fix few days ago on zonemaster.net. The logo might appear on Chrome now.

matsduf:
That is good.

PR #187, opened 2021-03-04
Resolves issue with logo in Chrome

matsduf:
Resolves issue #181.

The solution has been tested and verified in Chrome, Firefox and Safari.


pamasse (approved):

pnax (commented):
I found it weird that Chrome does not support SVG.
Looking at this, it looks like we are using a wrong MIME type `data:img/png` in the zonemaster_logo_2021_color.svg file.

If you update the occurrences of `data:img/png` to `data:image/png` in the svg file, the logo looks like it is properly rendered. Tested in Chromium and Firefox.

Updating the MIME type is maybe just a tweak, and going with PNG might be more sustainable. Should we move to PNG or stick to SVG ?

source : https://css-tricks.com/forums/topic/svg-css-background-image-not-showing-in-chrome/#post-244623
valid MIME types : https://www.iana.org/assignments/media-types/media-types.xhtml

matsduf:
If you compare the old 2020 logotype in SVG format with the new 2021 logotype in SVG format you can see that they quite different. Is the old one a true vector image and the new one just a bit map PNG image encapsulated in SVG? If so we might as well use the PNG image as this PR does.

We could also try to "order" a new logotype in SVG format, but created as vector image.

Please note that I am not sure about the format of the images.

pnax:
I think I have the same knowledge as you on this. So if this is not going backwards (I mean replace an old deliberate choice to use vector images) I agree to use a PNG file here.

matsduf:
> I think I have the same knowledge as you on this. So if this is not going backwards (I mean replace an old deliberate choice to use vector images) I agree to use a PNG file here.

It was @pamasse who wrote the current GUI. I do not think that using the SVG logotype has been an important decision. In the future we can go back to SVG, but then we should verify the logotype files stricter.

@pnax, if you press "approve" I can merge.

mattias-p (approved):

matsduf:
## Release testing for v2021.1
I have verified that latest version of Chrome displays the logotype under Windows, MacOS and Ubuntu.

Issue #189, opened 2021-03-19
Date format in history view


mattias-p:
Today the history view uses a date format that looks like this "3/19/21, 6:36 PM GMT+1". We should use ISO 8601 (with a SPACE instead of the T).

More precisely the format `YYYY-mm-dd HH:MM±hhmm` should be used.

matsduf:
This date-time stamp is for human consumption. I support the suggestion, and I suggest that we limit it to minutes, i.e. not seconds. 

A second question is time zone. One possibility is to have UTC only, another would have a configuration setting that sets the time zone. In both cases I think the time zone should be explicitly stated.

Today there is a time zone specification, but that is wrong. I executed at test at zonemaster.net at 14.05 UTC, and when the test is displayed it says "March 22, 2021 3:05 PM" without time zone, but in the history list "3/22/21, 2:05 PM GMT+1" which is 13.05 UTC.

If the browser signals time zone -- does it? -- then using that would be useful for the users. Still it should explicitly state the time zone.

mattias-p:
The exact format your suggesting is `YYYY-mm-dd HH:MM±hhmm`, right? I updated the description to that effect.

If you're getting the wrong date reported back to you that deserves a dedicated bug report.

I'm not familiar with the framework we're using for the GUI, but it looks to me like the time zone being used is whatever the client has configured as local time.

matsduf:
It is a standard, but is that a user friendly way of show time zone? Is it obvious for most people how `±hhmm` should be interpreted?

mattias-p:
I would say `YYYY-mm-dd HH:MM` is by far the most user friendly format. It's unambiguous with regard to field order. You can quickly get a sense of the distance between two adjacent dates.

So what are you suggesting?

matsduf:
I agree with you, `YYYY-mm-dd HH:MM` is very clear and user friendly. I was thinking of and discussing the time zone. `±hhmm` is not crystal clear.

mattias-p:
Ok, I get it. So how do you suggest we go forward?

matsduf:
The issue has been created, and now we have to see who can provide a solution.

pnax:
I agree with the proposal. I think we can use UTC for all displayed dates in the GUI.
And some pointers to the documentation :

pnax:
Closed by #191

PR #191, opened 2021-03-31
Consistent date format

pnax:
This updates the date format to `YYYY-mm-dd HH:MM UTC` everywhere (in the result page and the history).
This requires that the Backend sends the `creation_time` in UTC (which is not the case with SQLite yet, see https://github.com/zonemaster/zonemaster-backend/pull/732)

Angular documentation on date formats : https://angular.io/api/common/DatePipe

Addresses #189

### DIFF
--- a/src/app/components/history/history.component.html
+++ b/src/app/components/history/history.component.html
@@ -2,7 +2,7 @@
   <div class="row">
     <div *ngIf="history.length > 0;else noHistory" class="list-group w-100">
       <div *ngFor="let item of historyItems" class="list-group-item list-group-item-action list-group-item-{{item.color}}">
-        <a routerLink="/result/{{item.id}}"> {{ item.creation_time | date:'yyyy-MM-dd HH:mm' }} UTC </a>
+        <a routerLink="/result/{{item.id}}"> {{ item.creation_time | amFromUtc | date:'yyyy-MM-dd HH:mm zzzz' }}</a>
 
       </div>
     </div>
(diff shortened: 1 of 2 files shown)


mattias-p (dismissed):

mattias-p:
I don't have a clear view of how much sense UTC makes to people across the globe, but in any case this is a clear improvement.

pnax:
I think you are right, UTC might not be the easiest to grasp (as it still needs conversion for the user). I converted it to the local time. The format is :
```
YYYY-mm-dd HH:MM GMT±hhmm
```

This uses [`amLocal` from ngx-moment](https://github.com/urish/ngx-moment#amlocal-pipe).

The following screenshots are taken with my timezone set to Asia/Tehran (thanks for the tip ;) ) :

```
sudo timedatectl set-timezone Asia/Tehran
```

zm-date-info

zm-date-history

mattias-p (approved):
I like it!

matsduf (approved):
I feel GMT is outdated. UTC is what is used. But this is better than now.

pnax:
> I feel GMT is outdated. UTC is what is used.

To be honest this is the output from the DatePipe with the `zzzz` format.

matsduf:
> > I feel GMT is outdated. UTC is what is used.
> 
> To be honest this is the output from the DatePipe with the `zzzz` format.

And we can live with that.

matsduf:
The description of this PR is confusing since it target is changed further down.

matsduf:
## Release testing for v2021.1
Time is displayed as expected from this PR.

pnax:
> The description of this PR is confusing since it target is changed further down.

Good point. I updated the PR's description.

PR #174, opened 2021-01-28
Add norwegian (nb)

matsduf:
This replaces the part of #165 that adds Norwegian (nb) to GUI (ac535b1). It is the same change with the additional change that #169 introduced.

This PR should not be merged until #163 has been resolved.

PR #195, opened 2021-04-12
Updated e2e language tests

matsduf:
* Added test script for da (Danish) language
* Added test script for fi (Finnish) language
* Added test script for nb (Norwegian) language
* Renamed language test scripts to be FR05-xx to make it possible
  to expand with new language and keep them in sequence

Partially resolves request in https://github.com/zonemaster/zonemaster-gui/pull/174#pullrequestreview-633479423

### DIFF
--- a/e2e/FR05-fr.e2e-spec.ts
+++ b/e2e/FR05-fr.e2e-spec.ts
@@ -2,7 +2,7 @@ import { by, browser, element } from 'protractor';
 
 import { Utils } from './utils/app.utils';
 
-describe('Zonemaster test FR06-fr - [Supports French language]', () => {
+describe('Zonemaster test FR05-fr - [Supports French language]', () => {
 
   const utils = new Utils();
 
--- a/e2e/FR05-fr.e2e-spec.ts
+++ b/e2e/FR05-fr.e2e-spec.ts
@@ -2,7 +2,7 @@ import { by, browser, element } from 'protractor';
 
 import { Utils } from './utils/app.utils';
 
-describe('Zonemaster test FR06 - [Supports French language]', () => {
+describe('Zonemaster test FR06-fr - [Supports French language]', () => {
 
   const utils = new Utils();
 
--- a/docs/TranslationGuide.md
+++ b/docs/TranslationGuide.md
@@ -19,7 +19,9 @@ following languages with the attached language code:
 
 * `da` for Danish language
 * `en` for English language
+* `fi` for Finnish language
 * `fr` for French language
+* `nb` for Norwegian language
 * `sv` for Swedish language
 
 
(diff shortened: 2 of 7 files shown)


e2e/FR05-fr.e2e-spec.ts
```diff
@@ -2,7 +2,7 @@ import { by, browser, element } from 'protractor';
 
 import { Utils } from './utils/app.utils';
 
-describe('Zonemaster test FR06 - [Supports French language]', () => {
+describe('Zonemaster test FR06-fr - [Supports French language]', () => {
```
pnax, line 5: I think this should be `FR05-fr` to be aligned with the rest.

pnax (commented):
A small miss in tests renaming I think.
And a question : any idea why all files start with the `FR` prefix ?

e2e/FR05-fr.e2e-spec.ts
matsduf: Corrected.

matsduf:
@pnax  and @mattias-p, please review. I have also updated the translation instruction document.

matsduf:
> And a question : any idea why all files start with the `FR` prefix ?

No. I tried to understand the magic in the file names, but I found no information. I just tried with new files, and they were run.

repperille:
What about Functional Requirement (FR). All tests should test the requirements of a piece of code. Is there a requirement specification somewhere?

matsduf:
> What about Functional Requirement (FR). 

Yes, possibly.

> All tests should test the requirements of a piece of code. Is there a requirement specification somewhere?

Unfortunately no written specification for Zonemaster-GUI.

mattias-p:
> What about Functional Requirement (FR).

That's my guess too.

> Is there a requirement specification somewhere?

The specification is embedded in the test code. I.e. in the string arguments to `describe()` and `it()` and friends. This way of doing things comes from [Behavior Driven Development](https://en.wikipedia.org/wiki/Behavior-driven_development#Story_versus_specification).

pamasse:
> Unfortunately, there is no written specification for Zonemaster-GUI in Github.

Since this GUI was a new version of the perl GUI, we first based the tests on old requirements that I don't find anymore. 
But I'm pretty sure that we had a file with a list of requirements for the GUI.

pnax (approved):

pnax:
>> What about Functional Requirement (FR).

> That's my guess too.

Ok, thank you for your answers :)

mattias-p (approved):
I had an idea that it might be easier to maintain this if all languages were combined into a single functional requirement. I'm not familiar with the testing framework being used and I'm not sure if that would carry any negative consequences, but I do think it's a thing to consider.

Apart from that I think this looks good.

matsduf:
> Since this GUI was a new version of the perl GUI, we first based the tests on old requirements that I don't find anymore.
> But I'm pretty sure that we had a file with a list of requirements for the GUI.

The [Old Zonemaster-GUI repository] is still available (but should be deleted), and there is no requirement list for GUI either. We only have the tests in the (current) Zonemaster-GUI repository that indirectly define the requirements. We should have something else.

[Old Zonemaster-GUI repository]: https://github.com/dotse/zonemaster-gui

matsduf:
> I had an idea that it might be easier to maintain this if all languages were combined into a single functional requirement. I'm not familiar with the testing framework being used and I'm not sure if that would carry any negative consequences, but I do think it's a thing to consider.

One could consider FR05 (after my change) to be a single language requirement, but executed in separate files. Having them in separate files will make it easier to add, modify and delete languages.

PR #196, opened 2021-04-12
Removes dead code

matsduf:


mattias-p:
Looks good to me, but I'd prefer if @pamasse could review this.

pnax (approved):
Looks good to me. (and it seems like this code is here since the first commit so might be safe to remove from my point of view).

matsduf:
> Looks good to me. (and it seems like this code is here since the first commit so might be safe to remove from my point of view).

I have also consulted @pamasse for this issue.

matsduf:
> Looks good to me, but I'd prefer if @pamasse could review this.

I have already consulted @pamasse.

pamasse (approved):

PR #202, opened 2021-05-28
CentOS installation

mattias-p:
## Purpose

Some fixes and cleaning up.

## Changes

Installation of unzip is added to the CentOS installation. It's needed to extract the dist tarball.

The smoke test instruction was outdated in terms of what to look for to verify success. This PR fixes that.

Whitespace in a sample file is updated to use spaces instead of tabs.

## How to test this PR

This PR is tested implicitly as part of installation testing.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -35,7 +35,7 @@ This instruction covers the following operating systems:
 
 ```sh
 sudo yum update
-sudo yum install httpd
+sudo yum -y install httpd unzip
 ```
 
 
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -186,13 +186,16 @@ Use the procedure for installation on [Debian](#2-debian).
 
 ## Post-installation sanity check
 
-Point your browser at `http://localhost/` (or the address of the server where
-you installed Zonemaster Web GUI).
+Make sure Zonemaster-GUI is properly installed.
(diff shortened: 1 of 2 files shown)


docs/Installation.md
```diff
@@ -189,8 +189,10 @@ Use the procedure for installation on [Debian](#2-debian).
 Point your browser at `http://localhost/` (or the address of the server where
 you installed Zonemaster Web GUI).
 
-Verify that the Zonemaster Web GUI is shown with the " Version used " in its page footer.
-And when the mouse over this text, it display all version of the zonemaster stack (Backend, Engine and GUI).
+Verify that the Zonemaster Web GUI is shown with the text "Program versions" in
+its page footer.
+Verify that when you mouse over this text the versions of the following
+Zonemaster components are shown: Backend, Engine and GUI.
```
matsduf, line 195: I suggest the following to make the two tests stand out better:
```
* Verify that the Zonemaster Web GUI is shown with the text "Program versions" in its page footer. 
* Verify that when you mouse over this text the versions of the following Zonemaster components are shown: Backend, Engine and GUI.
```

mattias-p: Good idea! I'll make it a list.

matsduf (approved):

Issue #188, opened 2021-03-16
Copyright Footer : Year as a variable ?


benoitam:
Hi, 

I see that the footer is outdated : "Copyright © AFNIC and The Swedish Internet Foundation, 2018-2020. All Rights Reserved"
What about showing the current year only?

Thanks !
Ben

matsduf:
The year is set to 2021.

PR #209, opened 2021-06-03
Update copyright year to 2021

pnax:
## Purpose

The copyright statement is still 2020. This updates it to 2021.

## Context

Addresses #188 (well not exactly because the issue is about to dynamically setting the year).

## Changes

Update the footer component inner HTML.

## How to test this PR

Look that the copyright statement displays the current year.

## Final note

I am no jurist, but do we really need this copyright statement ?

### DIFF
--- a/src/app/components/footer/footer.component.html
+++ b/src/app/components/footer/footer.component.html
@@ -20,7 +20,7 @@
 
     <div class="row">
       <div class="col-md-12">
-        <p class="copyright text-muted small">Copyright &copy; AFNIC and The Swedish Internet Foundation, 2018-2020. All Rights Reserved</p>
+        <p class="copyright text-muted small">Copyright &copy; AFNIC and The Swedish Internet Foundation, 2018-2021. All Rights Reserved</p>
       </div>
     </div>
   </div>


matsduf:
I do not think means much if the year is correct or not.

pnax:
> Could we defer this to v2021.2?

From my point of view this is totally fine.

> I do not think means much if the year is correct or not.

I don't know either. I would even suggest to remove the whole copyright statement if we think it is not needed. We can keep something like `developed and maintained by Afnic and The Swedish Internet Foundation since 2013` (but this would be another PR if you all agree on that).

I'll leave this PR open for now. Feel free to comment on the need of the copyright statement (and if yes, what should go in it).

matsduf:
My proposal:
* Update this PR to remove year from this place. It is enough with the copyright statement.
* We have year in the [LICENSE] file. 
  * Ask lawyers if needed.
  * If it should be there we have to include in the release process to update it in all repositories, else remove it from all such files.

[LICENSE]: https://github.com/zonemaster/zonemaster-gui/blob/master/LICENSE

mattias-p (approved):

mattias-p:
What is it we're claiming copyright for anyway? This copyright statement is in HTML. Does the copyright statement cover the user interface? Maybe the text in the FAQ? It shows up on pages that show test results, but I doubt we want those to be covered by copyright.

matsduf (changes requested):
Deprecated by #215

pnax:
Not relevant anymore, since #215 removes the year.

PR #214, opened 2021-06-04
Updates installation instruction for FreeBSD

matsduf:
## Purpose

Unify the instructions with those for Engine (for FreeBSD) and to make copy-and-past easier,

## Changes

* Adds the same pkg setting as in Zonemaster-Engine.
* Updates to make copy-and-past easier.

## How to test this PR

Perform an installation on FreeBSD.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -38,7 +38,6 @@ sudo yum update
 sudo yum -y install httpd unzip
 ```
 
-
 #### Install Zonemaster Web GUI
 
 ```sh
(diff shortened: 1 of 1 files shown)


matsduf:
@pnax and @mattias-p, please review.

pnax (approved):

docs/Installation.md
```diff
@@ -127,32 +123,65 @@ sudo systemctl reload apache2
...
+
+#### Update list of package repositories
+Create the file `/usr/local/etc/pkg/repos/FreeBSD.conf` with the following
+content, unless it is already updated:
+
+```
+FreeBSD: {
+url: "pkg+http://pkg.FreeBSD.org/${ABI}/latest",
+}
+```
```
mattias-p, line 138: What would happen if we removed this? (From here and from Engine as well).

matsduf: Older packages would be installed. I added this when there was an inconsistency between the packages in "monthly", which is the default.

matsduf:
## Testing v2021.2
LGTM.

PR #216, opened 2021-06-10
Makes DS hash algorithm number explicit

matsduf:
## Purpose

The hash algorithm for DS is often displayed as a number, e.g. with `dig`. In the GUI when you do a undelegated test and select DS parameter the hash algorithm is selected by name. It should be possible to start with the algorithm code.

## Changes

This RP updates the select list so that the hash algorithm code is also shown. The solution is the same as for the DNSKEY algorithm select list.

## How to test this PR

Verify that correct algorithm is selected.

### DIFF
--- a/src/app/components/form/form.component.html
+++ b/src/app/components/form/form.component.html
@@ -135,10 +135,10 @@
                 </select>
                 <select formControlName="digtype" class="form-control" name="form.digtype">
                   <option [value]="-1" selected disabled hidden> {{'Digest type' | translate}} </option>
-                  <option [value]="1">SHA-1</option>
-                  <option [value]="2">SHA-256</option>
-                  <option [value]="3">GOST R 34.11-94</option>
-                  <option [value]="4">SHA-384</option>
+                  <option [value]="1">1 - SHA-1</option>
+                  <option [value]="2">2 - SHA-256</option>
+                  <option [value]="3">3 - GOST R 34.11-94</option>
+                  <option [value]="4">4 - SHA-384</option>
                 </select>
                 <input formControlName="digest" class="form-control" name="digest" placeholder="{{'Digest' | translate}}">
                 <button


src/app/components/form/form.component.html
```diff
@@ -135,10 +135,10 @@ <h4 class="d-inline">{{'DS_RECORDS'|translate}}</h4>
...
                 <select formControlName="digtype" class="form-control" name="form.digtype">
                   <option [value]="-1" selected disabled hidden> {{'Digest type' | translate}} </option>
-                  <option [value]="1">SHA-1</option>
-                  <option [value]="2">SHA-256</option>
-                  <option [value]="3">GOST R 34.11-94</option>
-                  <option [value]="4">SHA-384</option>
+                  <option [value]="1">1 - SHA-1</option>
+                  <option [value]="2">2 - SHA-256</option>
+                  <option [value]="3">3 - GOST R 34.11-94</option>
+                  <option [value]="4">4 - SHA-384</option>
```
mattias-p, line 141: Have you considered using the notation `SHA-1 (algo 1)`? I think that would be clearer.

mattias-p (approved):

src/app/components/form/form.component.html
matsduf: I modeled it after how the DNSKEY algorithms in the same file are specified, see lines 123-134.

In the DNS records the algorithms are given as numbers. If you write them as  "(algo 1)" at the end then I think it will be less clear.

image

mattias-p: I agree that format should be consistent across both lists. If both lists used the (algo n) format, that might be clearer.

matsduf: In the present format, the algorithm is listed below each other, with your proposal, it would be less clear.

Current and current proposal:
```
1 - RSAMD5
3 - DSA
5 - RSASHA1
6 - DSA-NSEC3-SHA1
7 - RSASHA1-NSEC3-SHA1
8 - RSASHA256
10 - RSASHA512
12 - ECC-GOST
13 - ECDSAP256SHA256
14 - ECDSAP384SHA384
15 - ED25519
16 - ED448
```
```
1 - SHA-1
2 - SHA-256
3 - GOST R 34.11-94
4 - SHA-384
```

New proposal:
```
RSAMD5 (ALGO 1)
DSA (ALGO 3)
RSASHA1 (ALGO 5)
DSA-NSEC3-SHA1 (ALGO 6)
RSASHA1-NSEC3-SHA1 (ALGO 7)
RSASHA256 (ALGO 8)
RSASHA512 (ALGO 10)
ECC-GOST (ALGO 12)
ECDSAP256SHA256 (ALGO 13)
ECDSAP384SHA384 (ALGO 14)
ED25519 (ALGO 15)
ED448 (ALGO 16)
```
```
SHA-1 (ALGO 1)
SHA-256 (ALGO 2)
GOST R 34.11-94 (ALGO 3)
SHA-384 (ALGO 4)
```

mattias-p: You do have a point that the alignment is neater with the algorithm numbers on the left. And I agree that my suggestion is quite terrible when you change "algo" to be in upper case.

But I'll try to state my point more clearly. I do agree that it's a good idea to spell out the algorithm numbers. What I'm questioning is whether it's obvious to non-experts that the integers are in fact algorithm numbers. And I tried to suggest a way to clarify that to our users who might be DNSEC neophytes. Maybe there are even better ways to present such a clarification, or maybe it's not needed. But in any case these are my reasons.

matsduf: > You do have a point that the alignment is neater with the algorithm numbers on the left. And I agree that my suggestion is quite terrible when you change "algo" to be in upper case.

I am sorry. My mistake. I have now updated above to lower case "algo".

> But I'll try to state my point more clearly. I do agree that it's a good idea to spell out the algorithm numbers. What I'm questioning is whether it's obvious to non-experts that the integers are in fact algorithm numbers. And I tried to suggest a way to clarify that to our users who might be DNSEC neophytes. Maybe there are even better ways to present such a clarification, or maybe it's not needed. But in any case these are my reasons.

I agree that documentation is lacking. There is for some reason no lead text in those two fields (see #226) and we could also add some documentation for the fields that opened when you press the "alternative" button.

matsduf: With PR #227 it looks like the screen shot. What do you think, @mattias-p?

image

mattias-p: I think this is a clear improvement.

It's still a bit cryptic for DNSSEC neophytes though. But if the price of making it better for DNSSEC neophytes is to make it worse for those who know DNSSEC, I don't think it's worth it.

pnax (approved):

pnax:
## Testing v2021.2

lgtm

Issue #217, opened 2021-06-10
Make IP option dynamic


blacksponge:
I have installed zonemaster on a IPv4 only server so I disabled the IPv6 option in the profile but if I understand correctly the GUI default to true on both IPv4 and IPv6 without checking the value in profile. This cause the test to be executed for both each time, resulting in connectivity issue (https://zonemaster.ppsfleet.navy/result/82f4deea1693fc5b). I suggest to add a method in the backend so the client can fetch the needed settings from the profiles and fill the GUI's defaults according to those values.

matsduf:
When the test is started the RPC call have three alternatives for `IPv4` and `IPv6`:

* true
* false
* (unset)

Both `true` and `false` will override what the profile says. If left unset then the setting in the profile will rule.

GUI will always set those to `true` or `false`, never unset. A reasonable change would be that GUI by default leaves those unset, and that the user can choose to set it to `true` or `false`.

```
IPv4: [x] default  [ ] yes  [ ] no  ||  IPv6: [x] default  [ ] yes  [ ] no
```

blacksponge:
This is quick fix indeed, the drawback I see is that the user doesn't know what the default is and that we may end up in situation where both ipv4 and ipv6 are disabled. I still implemented it here #218 as this is still an improvement.

matsduf:
I have given this issue a second thought. I do not think that the three-way selection on both IPv4 and IPv6 that I proposed above in https://github.com/zonemaster/zonemaster-gui/issues/217#issuecomment-858662304 is a good idea. It increases the complexity for the users to solve very rare use cases. With that I do not think that the present solution is good either.

For most users the GUI sits in front of a backend with both IPv4 and IPv6 connection and with both IPv4 and IPv6 being enabled in the profile. Turning on IPv4 or IPv6 has then no meaning since that is the default. Turning off IPv4 or IPv6 (but not both) could be meaningful for troubleshooting. What we need are buttons to be able to disable IPv4 or IPv6.

I rare situations we have a GUI in front of a backend with IPv4 only or IPv6 only. Then the other network can be disabled in the profile. In such a case buttons that can turn the networks off have no meaning, but they do not interfere with the profile setting, as the current solution.

The use cases where someone wants to set to `true` where it is `false` in the profile are very rare. They could be supported by running CLI on local machine instead.

[...]

PR #218, opened 2021-06-10
add default option for protocol setting

blacksponge:
## Purpose

This way the default settings on the GUI will use the default on the profile with the ability to override those settings.

## Context

This is a proposal following discussion on #217.

## How to test this PR

* Using the default setting on a dual stack backend should not change anything.
* Disabling one protocol should disable the tests using that protocol, could be checked in test results.
* Testing a domain using backend that only support one protocol (ex. zonemaster.ppsfleet.navy ) with the default settings should only trigger the tests of the supported protocol.

Screenshot 2021-06-14 at 16-36-46 Zonemaster


matsduf (changes requested):
Your proposed update is not bad, and it follows my first suggestion. But I think we should give it a second thought. See my updated suggestion in https://github.com/zonemaster/zonemaster-gui/issues/217#issuecomment-859348817, which will also solve the original issue.

blacksponge:
Updated version:
Screenshot 2021-06-14 at 16-36-46 Zonemaster

matsduf (commented):
It looks fine. I will run a check before I formally approve it. Please add
```
"Disable IPv6": "Inaktivera IPv6",
"Disable IPv4": "Inaktivera IPv4"
```
in sv.json

src/environments/environment.ts
```diff
@@ -6,6 +6,8 @@
 export const environment = {
   production: false,
   apiEndpoint: 'https://zonemaster.net/api',
+  //apiEndpoint: 'https://zonemaster.ppsfleet.navy/api',
```
matsduf, line 9: What is it? What is it for? How and when is it used?

blacksponge: Sorry, I pushed too much stuff, will revert.

matsduf (dismissed):
This works as expected when tested. I approve this. Either we should have the translation in the remaining languages in this PR or in a separate one.

matsduf:
@samisalmensuo, @repperille and @haarbo!

Screenshot 2021-06-14 at 16-36-46 Zonemaster

Could you provide the translations of "Disable IPv6" and "Disable IPv4", respectively, into Finnish, Norweigian and Danish, respectively? Just add a comment here.

haarbo:
Disable is "Deaktiver" in Danish.

matsduf:
> Disable is "Deaktiver" in Danish.

Thanks!

matsduf (dismissed):

src/assets/i18n/fr.json
```diff
@@ -48,5 +48,7 @@
 	"What is an undelegated domain test?": "C'est quoi un test sur un nom de domaine non délégué?",
 	"Zonemaster Backend is not available": "Zonemaster Backend n'est pas disponible",
 	"Profile": "Profil",
-	"Copyright © AFNIC and The Swedish Internet Foundation. All Rights Reserved": "© AFNIC et The Swedish Internet Foundation. Tous droits réservés"
+	"Copyright © AFNIC and The Swedish Internet Foundation. All Rights Reserved": "© AFNIC et The Swedish Internet Foundation. Tous droits réservés",
+	"Disable IPv6": "Désactivé IPv6",
+	"Disable IPv4": "Désactivé IPv4"
```
pnax, line 53: I would go with the infinitive form -> `Désactiver`.

pnax (commented):
It looks like when a test finishes, the value is not properly reset.

Using the GUI from this PR (commit da6ebb3)
1. disable IPv4
2. run a test (for instance for afnic.fr)
3. wait for the test to finish and verify that IPv4 was disabled (OK)
4. open the « Options » menu again
5. both button are disabled
6. run a test (for instance for iis.se)
7. wait for the test to finish and see that IPv4 was disabled (not OK)

Either keep the button to disable IPv4 engaged or reset the value sent to the Backend.

src/assets/i18n/fr.json
blacksponge: Fixed.

blacksponge:
> It looks like when a test finishes, the value is not properly reset.

Good catch, I will fix that as well.

matsduf:
I cannot build a release.
```
ERROR in src/app/components/form/form.component.html:13:33 - error TS2339: Property 'domain' does not exist on type '{}'.

13                    [(ngModel)]="form.domain"
                                   ~~~~~~~~~~~

  src/app/components/form/form.component.ts:13:16
    13   templateUrl: './form.component.html',
                      ~~~~~~~~~~~~~~~~~~~~~~~
    Error occurs in the template of component FormComponent.
src/app/components/form/form.component.html:13:45 - error TS2339: Property 'domain' does not exist on type '{}'.

13                    [(ngModel)]="form.domain"
                                               
14                    (keypress)="($event.which === 13)? runDomainCheck() : 0"
   ~~~~~

  src/app/components/form/form.component.ts:13:16
    13   templateUrl: './form.component.html',
                      ~~~~~~~~~~~~~~~~~~~~~~~
    Error occurs in the template of component FormComponent.
src/app/components/form/form.component.html:18:62 - error TS2339: Property 'domain' does not exist on type '{}'.

18               <p class="reset-form d-none d-sm-block" *ngIf="form.domain.length > 0 && !showProgressBar" >
                                                                ~~~~~~~~~~~

  src/app/components/form/form.component.ts:13:16
    13   templateUrl: './form.component.html',
                      ~~~~~~~~~~~~~~~~~~~~~~~
    Error occurs in the template of component FormComponent.
```

blacksponge:
Can you retry?

matsduf:
> Can you retry?

Now I could build.

matsduf (dismissed):
I have tested and I think it behaves as it should. This is an improvement over the current solution.

matsduf:
The radio buttons here look wrong:
image

matsduf (changes requested):
I suggest that the radio buttons about IPv4 and IPv6 are removed. In the report it says if any of them are disabled, and the buttons can only say it was requested to be disabled. It has no control of the profile. Most users of the GUI do not care about IPv4 and IPv6.

blacksponge:
> The radio buttons here look wrong:

What do you mean?

Also, I think that the radio buttons in the result are beyond the scope of this MR.

matsduf:
> Also, I think that the radio buttons in the result are beyond the scope of this MR.

This PR has changed how the radio buttons for IPv4 och IPv6 are displayed. The screen shot below shows how it looks like on zonemaster.net if you enable IPv4 and disable IPv6. IPv4 is green and IPv6 is red. With the updated code then the radio buttons are always red.

image

One solution is to remove them.

blacksponge:
Oh okay, that probably comes from the fact that the ipv4 and ipv6 params are not explicitly stated in the request any more. Ideally the backend should send back the actual parameters that were used in the tests not just the one filled by the user (it will help on tests reusability too). But, yes, as a workaround they could be removed.

blacksponge:
Acutally if I understand correctly, https://github.com/zonemaster/zonemaster-backend/pull/798 might solve that issue.

matsduf:
> Acutally if I understand correctly, [zonemaster/zonemaster-backend#798](https://github.com/zonemaster/zonemaster-backend/pull/798) might solve that issue.

No, it has nothing to do with the Backend. Those buttons are set by the GUI based on the selection by the user.

blacksponge:
Those button have nothing to do with the input form, they are set in the result component here
https://github.com/blacksponge/zonemaster-gui/blob/ip-option-default/src/app/components/result/result.component.html#L39
based on the data received from the `get_test_results` call here
https://github.com/blacksponge/zonemaster-gui/blob/ip-option-default/src/app/components/result/result.component.ts#L141 .

The `params` field in the result do not have any ipv4/ipv6 key as the input has not been normalized, hence resulting in `undefined` which is evaluated to false when interpreted in the template.

matsduf:
> Those button have nothing to do with the input form, they are set in the result component here
(...)
> The `params` field in the result do not have any ipv4/ipv6 key as the input has not been normalized, hence resulting in `undefined` which is evaluated to false when interpreted in the template.

Thank you for clarifying. The change in this PR will make the ipv4/ipv6 keys being unset in the request. As far as I understand the value of the ipv4/ipv6 keys are copied from the request to the `get_test_results` response. It means that this FQ has created a bug concerning the ipv4/ipv6 buttons. I do not think it is acceptable to leave it as is (always being red), but I think it is acceptable to just remove the buttons because I do not think they add any value.

mattias-p:
Aren't the markings there to signify what protocols were used for this particular test? I would think that's a good thing to know when you're trying to understand the results.

matsduf:
If a protocol was disabled, there is a message in Basic. With this PR and the design of the Backend, the GUI has no access to the setting in the profile.

matsduf:
image

blacksponge:
>  As far as I understand the value of the ipv4/ipv6 keys are copied from the request to the get_test_results response

With zonemaster/zonemaster-backend#798 it seems that they will be set to the profile value if left unset. 
https://github.com/zonemaster/zonemaster-backend/pull/798/files#diff-7a0828c7a5501d04ebe0287191fe90c840c0cdb0a6670c77ff7e4ac257decec7R144-R145

matsduf:
No, the method `generate_deterministic_hash` in  zonemaster/zonemaster-backend#798 will not change the value of the ipv4/ipv6 parameters, only create a local hash with effective values, and then create a hash. -- And that is good because that would change the behavior of the API.

blacksponge:
But doesn't it also update the `encoded_params` hash that is then inserted into the db and is later the one retrieved by `test_results`? Or am I missing something?

matsduf:
The deterministic_hash is a MD5 hash inserted in the database, but the function does not change the parameters themselves.

blacksponge:
I am sorry but I don't understand how it doesn't change. From what I see,

* a new hash `to_encode` is created to compute the determinist hash [ref](https://github.com/pnax/zonemaster-backend/blob/deterministic-hash/lib/Zonemaster/Backend/DB.pm#L141),
* this hash is json encoded and returned to the caller along with the md5 hash [ref](https://github.com/pnax/zonemaster-backend/blob/deterministic-hash/lib/Zonemaster/Backend/DB.pm#L157)
* `create_new_test` use the returned encoded parameters [ref](https://github.com/pnax/zonemaster-backend/blob/deterministic-hash/lib/Zonemaster/Backend/DB/MySQL.pm#L129) to create the new test in the db [ref](https://github.com/pnax/zonemaster-backend/blob/deterministic-hash/lib/Zonemaster/Backend/DB/MySQL.pm#L158)
* those same parameters are used for the test  [ref](https://github.com/pnax/zonemaster-backend/blob/deterministic-hash/lib/Zonemaster/Backend/DB/MySQL.pm#L195) and return by `test_results` [ref](https://github.com/pnax/zonemaster-backend/blob/deterministic-hash/lib/Zonemaster/Backend/DB/MySQL.pm#L218)

The parameters are not modified in place but a new hash is replacing them to it is the same. We can probably discuss that at another place but I really feel like I am missing something because if it doesn't change I can't seem to see why.

matsduf:
> I am sorry but I don't understand how it doesn't change. From what I see,

You are right, but that is just a draft that have not approved yet. Either we have to wait with this PR, or we have to resolve the bug within this PR. I suggest the latter.

blacksponge:
Okay I'll do that next week then

matsduf:
@mattias-p and @pnax!  Please express your opinion.

https://github.com/zonemaster/zonemaster-gui/pull/218/commits/86ab70a1227e38eb002783b66aba3e13b3f1cd39 removes the radio buttons that display of transport protocol (IPv4/IPv6) since with the update that will not be correct.

If we change the API  as suggested in  zonemaster/zonemaster-backend#798 then the buttons could be kept, but this PR cannot be merged until we have the change in the API.

Is displaying the transport protocol so important? In the messages you can always see when a protocol has been turned off. I do not think that most GUI users care about transport protocol. A think we can remove the radio buttons from the GUI (but the change in the API is maybe still good).

matsduf (approved):
This is fine. Can @mattias-p and @pnax review too?

mattias-p:
I think displaying the query parameters on the result page is a great feature. Ideally we should have a clear section above the 
result messages where we display all the (normalized) query parameters in an structured way. That's out of scope for this PR, of course.

I don't think the result messages is a great place to communicate query parameters, but it's better than nothing. A couple of check boxes below a settings button is so great either. I don't mind having the check boxes removed, but then maybe we should also make an effort to get query parameters onto the result page in a more structured way.

e2e/FR12.e2e-spec.ts
```diff
@@ -10,12 +10,12 @@ describe('Zonemaster test FR12 - [The simple view should support an advanced vie
...
-    expect(element(by.css('label[for="protocol_ipv6"]')).isPresent()).toBe(false);
+    expect(element(by.css('label[for="disable_protocol_ipv4"]')).isPresent()).toBe(false);
+    expect(element(by.css('label[for="disable_protocol_ipv6"]')).isPresent()).toBe(false);
   });
   it('should have [IPv4 checkbox] & [IPv6 checkbox] visible', () => {
     element(by.css('.switch')).click();
-    expect(element(by.css('label[for="protocol_ipv4"]')).isPresent()).toBe(true);
-    expect(element(by.css('label[for="protocol_ipv6"]')).isPresent()).toBe(true);
+    expect(element(by.css('label[for="disable_protocol_ipv4"]')).isPresent()).toBe(true);
+    expect(element(by.css('label[for="disable_protocol_ipv6"]')).isPresent()).toBe(true);
```
mattias-p, line 19: Maybe add "Disable" to the it() clauses?

src/app/components/form/form.component.html
```diff
@@ -70,16 +70,15 @@ <h1>{{'Domain name'|translate}}</h1>
       <div *ngIf="is_advanced_options_enabled">
              <span [formGroup]="checkboxForm" class=" half-with">
               <span class="options d-flex" [formArrayName]="'items'" [class.invalid]="!checkboxForm['controls'].selectedItems.valid">
-                <span *ngFor="let protocol of checkboxForm['controls'].items['controls']; let i = index;"
+                <div *ngFor="let protocol of checkboxForm['controls'].items['controls']; let i = index;"
```
mattias-p, line 73: Is it proper nowadays to have div-tags inside of span-tags?

mattias-p (commented):
Looks real good to me. I have a suggestion and a question.

src/app/components/form/form.component.html
blacksponge: It is not ideal semantically speaking but would work without issue, I still revert it back to span

e2e/FR12.e2e-spec.ts
blacksponge: This is done.

pnax:
I agree with @mattias-p about having an easy way to see query parameters. This is fine for me to remove the buttons in the result page for now. But I think it would be nice to have them back when we'll decide to display the query parameters.

blacksponge:
So can this be merged?

matsduf:
> So can this be merged?

I will review.

pnax (commented):
I tested it and it seems ok.
Before formally approving the PR, could you update the description with the right screenshot to avoid reading the comments and also add a small text on how to test this PR ?

matsduf:
> @samisalmensuo, @repperille and @haarbo!
> 
> Screenshot 2021-06-14 at 16-36-46 Zonemaster
> 
> Could you provide the translations of "Disable IPv6" and "Disable IPv4", respectively, into Finnish, Norweigian and Danish, respectively? Just add a comment here.

@samisalmensuo and @repperille, could you provide the missing translation to Finnish and Norwegian, respectively?

matsduf (approved):
This looks fine. It behaves as expected when tested.

Having the translation into Finnish and Norwegian is not a show stopper for this PR, as it can be added by a separate PR afterwards. If the translations are not available when this is merged, please create an issue for that.

matsduf:
@pnax and @mattias-p, please review.

matsduf:
@blacksponge, I think you can merge.

pnax (approved):

mattias-p:
I've tested this for v2021.2 and it LGTM.

PR #220, opened 2021-06-14
pin webdriver version

### LINKED ISSUE #210: Travis fails
Since recently Travis is failing. It might be because of the Chromedriver that uses a new version of Chrome.
See PR #207 where this has been discovered, especially these comments https://github.com/zonemaster/zonemaster-gui/pull/207#issuecomment-853144960 and https://github.com/zonemaster/zonemaster-gui/pull/207#issuecomment-853152984

Maybe the error is somewhere else... I have not much clue for now, but it would be nice to see Travis happy again ;)

blacksponge:
Fix #210

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,9 @@ sudo: false
 addons:
   chrome: stable
 
-
 install:
   - npm install
+  - node ./node_modules/.bin/webdriver-manager update --versions.chrome=90.0.4430.24
 
 script:
-  - xvfb-run npm run e2e
+  - xvfb-run npm run ng e2e -- --webdriver-update=false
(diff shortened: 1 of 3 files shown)


mattias-p:
Does this mean that Chrome could introduce changes that are incompatible with Zonemaster GUI without us noticing?

blacksponge:
Unfortunately, yes, before the issue with protractor/chrome-driver is fixed we can't update to the latest stable without the tests breaking. So I would suggest to pin the version temporally so that we can at least catch regression in our code base.

mattias-p:
Maybe we could have an issue marked with milestone v2021.2 reminding us to revert the pinning?

blacksponge:
Yes, we can track this directly in #210, or create an other one.

Another workaround would be to use the `browser.executeScript` API to fetch the attributes like this

```js
getAttribute(element, attribute) {
    return browser.executeScript(`return arguments[0][arguments[1]];`, element, attribute);
}
```

I'm not sure what the best thing to do would be, either way it should only be temporary.

matsduf:
@pamasse, can you take a look at this?

pamasse (approved):

matsduf (approved):

matsduf:
@blacksponge, please merge.

PR #221, opened 2021-06-15
update client version in mock

blacksponge:
## Purpose

As the client version changed, the `start_domain_test` request was not mocked resulting on all subsequent request to not be mocked as well.

## Context

The e2e test FR22 start a test, all request to the API are intercepted and the ones that matches (equality check) the requests in mock.interceptor.ts are mocked. As the client version is part of the request not having the same will result in the requests not being mocked during the tests, which can lead to inconsistency.

## Changes

Use the current client version/id in the interceptor.

## How to test this PR

`npm run ng e2e -- --webdriver-update=false --specs=e2e/FR22.e2e-spec.ts`

### DIFF
--- a/src/app/interceptors/mock.interceptor.ts
+++ b/src/app/interceptors/mock.interceptor.ts
@@ -1,6 +1,8 @@
 import { Injectable, Injector } from '@angular/core';
 import { HttpEvent, HttpHandler, HttpInterceptor, HttpRequest, HttpResponse } from '@angular/common/http';
 import { Observable, of } from 'rxjs';
+import { environment } from '../../environments/environment';
+
 
 const urls = [
   {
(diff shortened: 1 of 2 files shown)


src/app/interceptors/mock.interceptor.ts
```diff
@@ -8,7 +8,7 @@ const urls = [
     body: {'jsonrpc': '2.0', 'id': 1572254767685, 'method': 'start_domain_test', 'params':
       {
         'ipv4': true, 'ipv6': true, 'profile': 'default', 'domain': 'afNiC.Fr',
-        'ds_info': [], 'nameservers': [], 'client_version': '3.1.0', 'client_id': 'Zonemaster GUI'
+        'ds_info': [], 'nameservers': [], 'client_version': '3.3.0', 'client_id': 'Zonemaster GUI'
```
matsduf, line 11: When is this call run? Mocking of what?

If this "client version" should be kept in sync with the release version of Zonemaster GUI then we should update the release instructions too. Else we could just set version "1".

matsduf:
@blacksponge, when you created this PR, didn't you get the template in the description?

src/app/interceptors/mock.interceptor.ts
blacksponge: I updated the description.

matsduf (approved):

Issue #213, opened 2021-06-04
Add full language at "mouse over"


matsduf:
When starting the GUI, the right corner has the language selection (here I disregard narrow windows):

<img width="236" alt="image" src="https://user-images.githubusercontent.com/5430733/120823462-bede5300-c557-11eb-882a-5d9f54ffcfb4.png">

Add the full language name when the mouse is moved over the language code.

blacksponge:
resolved by #222

Issue #212, opened 2021-06-04
Make selected language stick out


matsduf:
When starting the GUI, the right corner has the language selection (here I disregard narrow windows):

<img width="236" alt="image" src="https://user-images.githubusercontent.com/5430733/120822962-3495ef00-c557-11eb-92b7-b5eae0554db6.png">

You cannot see what language has been selected. Make the selected language stick out. It could be a combination of increasing the size of the letters, making the letters bold face, changing the backround or something else.

blacksponge:
resolved by #222

PR #222, opened 2021-06-15
Improve language selection nav

blacksponge:
## Purpose

Improve language selection nav on wide screens.

Some tests (FR16) are failing because the lang has not been changed when the test is conducted. Tries to wait (after the selected language has changed) for the translation to be effective before continuing. (replace #225)

## Context

Addresses #212, #213 and #223.

## Changes

Screenshot_20210615_113053

## How to test this PR

* Change languages and hover links.
* `npm run ng e2e -- --webdriver-update=false --specs=e2e/FR16.e2e-spec.ts `
* `npm run ng e2e -- --webdriver-update=false --specs='e2e/FR05-*.ts'`

### DIFF
--- a/e2e/FR16.e2e-spec.ts
+++ b/e2e/FR16.e2e-spec.ts
@@ -19,14 +19,11 @@ describe('Zonemaster test FR16 - [The advanced view should have a text describin
   it('should have a description text in multi languages', async () => {
     await utils.setLang('en');
     expect(element(by.css('.alert.alert-info')).element(by.css('a')).getText()).toContain('undelegated');
-    expect(element(by.css('nav div.lang a.selected')).getAttribute('lang')).toBe('en');
 
     await utils.setLang('fr');
     expect(element(by.css('.alert.alert-info')).element(by.css('a')).getText()).toContain('non délégué');
-    expect(element(by.css('nav div.lang a.selected')).getAttribute('lang')).toBe('fr');
 
     await utils.setLang('sv');
     expect(element(by.css('.alert.alert-info')).element(by.css('a')).getText()).toContain('odelegerat domäntest');
-    expect(element(by.css('nav div.lang a.selected')).getAttribute('lang')).toBe('sv');
   });
 });
--- a/e2e/FR16.e2e-spec.ts
+++ b/e2e/FR16.e2e-spec.ts
@@ -4,10 +4,10 @@ import { Utils } from './utils/app.utils';
 
 describe('Zonemaster test FR16 - [The advanced view should have a text describing what undelegated means?]', () => {
   const utils = new Utils();
-  beforeAll(() => {
-    utils.goToHome();
-    utils.setLang('en');
-    utils.activeOptions();
+  beforeAll(async () => {
+    await utils.goToHome();
+    await utils.setLang('en');
+    await utils.activeOptions();
   });
 
   it('should have a link to the proper faq answer', () => {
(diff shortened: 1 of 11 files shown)


matsduf (dismissed):
This looks fine! I have tested and it works nicely.

mattias-p:
It looks nice visually.

Maybe the functional tests should be updated to make sure this keeps working?

blacksponge:
I've updated the tests and implemented the changes discussed in #225

e2e/FR16.e2e-spec.ts
```diff
@@ -16,12 +16,17 @@ describe('Zonemaster test FR16 - [The advanced view should have a text describin
     expect(element(by.css('.alert.alert-info')).element(by.css('a')).getAttribute('fragment')).toBe('q12');
   });
 
-  it('should have a description text in multi languages', () => {
-    utils.setLang('en');
+  it('should have a description text in multi languages', async () => {
+    await utils.setLang('en');
     expect(element(by.css('.alert.alert-info')).element(by.css('a')).getText()).toContain('undelegated');
-    utils.setLang('fr');
+    expect(element(by.css('nav div.lang a.selected')).getAttribute('lang')).toBe('en');
```
mattias-p, line 22: What does element() do when the locator matches multiple elements? In case it doesn't fail, maybe we should count the number of selected elements too, to make sure there is only one?

mattias-p (approved):
Looks really good to me.

pnax:
I tried it, and the selected language is in blue for me :

zonemaster-gui-nav

And previous selection stays blue :

zonemaster-gui-nav-select

blacksponge:
> I tried it, and the selected language is in blue for me

Hum, I'll have a look 🤔

e2e/FR16.e2e-spec.ts
blacksponge: It was only emitting a warning, I've updated the tests accordingly, and also moved them to FR05-* which seemed more appropriate.

blacksponge:
@pnax I tried something but I'm having a hard time reproducing your case, can you tell me if that's better? Also does the text become underlined when you hover your mouse over the linked?

matsduf:
I have just installed the code from this PR with the latest commit (a3e1dbb) and I see no coloring and things are updated as they should. (I prefer that it is black as I see it.) I have tested on both Firefox och Chrome.

image

And now I switch to Fi:

image

matsduf:
@pnax, how did you test it?

pnax:
@blacksponge 
> I tried something but I'm having a hard time reproducing your case, can you tell me if that's better?

Yes, now the links are in black.

> Also does the text become underlined when you hover your mouse over the links?

Nope, when my mouse goes over another language, the text is not underlined. The text is underlined only for the selected language.

@matsduf 
> how did you test it?

Using `npm` on the PR branch, and TBH I struggled a little bit with npm (I also tried to run the tests), maybe I made something that broke the way the GUI is built.

matsduf:
> > how did you test it?
> 
> Using `npm` on the PR branch, and TBH I struggled a little bit with npm (I also tried to run the tests), maybe I made something that broke the way the GUI is built. Then I used Firefox to access the GUI.

I also checked the PR branch out, and then I ran `npm install && npm run release` and installed the resulting zip file on another VPS as specified in the installation instructions. I ignored anything that `npm` complained about.

pnax:
Well, my bad, I think I mixed up few things while testing. I even tried using commit e1d8620 (parent of a3e1dbb), and this seems to work as expected.

matsduf:
> I've updated the tests and implemented the changes discussed in #225

Please copy this into the description.

pnax:
> Also does the text become underlined when you hover your mouse over the links?

And just to point out that the text is not underlined when hovered (I don't know if this is the expected behavior).

matsduf (approved):
This is a good improvement!

blacksponge:
> And just to point out that the text is not underlined when hovered (I don't know if this is the expected behavior).
@pnax it is, the colour of your links make me though that they were styled like normal links (blue with underline on hover) so I was wondering if that could also be the case for you.

pnax (approved):

mattias-p (approved):

matsduf:
@mattias-p, please re-review.

matsduf:
@blacksponge, please merge.

pnax:
## Testing v2021.2

LGTM

PR #224, opened 2021-06-15
Omit empty ip addresses in param

### LINKED ISSUE #176: Omit empty IP addresses in calls to start_domian_test
As documented in https://github.com/zonemaster/zonemaster/issues/925#issuecomment-771744932, when the IP field is empty the GUI sends its "start_domain_test" with an empty string in the "ip" parameter. While the problem was mitigated with a quick fix in https://github.com/zonemaster/zonemaster-backend/pull/705, a better long term fix is to make GUI omit the "ip" parameter for IP fields that are empty.

blacksponge:
Fix #176


matsduf (approved):

mattias-p:

PR #223, opened 2021-06-15
Added to make Travis test. Not to be kept.

matsduf:
## Purpose

Trigger Travis to test. Not to be merged.

## Context

#214 and #216 unexpectedly fails.

PR #225, opened 2021-06-15
await setLang for FR16

blacksponge:
## Purpose

Some tests (FR16) fails, it looks like the lang has not been changed yet. Tries to await.

## Context

Following #223

### DIFF
--- a/e2e/FR16.e2e-spec.ts
+++ b/e2e/FR16.e2e-spec.ts
@@ -18,10 +18,13 @@ describe('Zonemaster test FR16 - [The advanced view should have a text describin
 
   it('should have a description text in multi languages', async () => {
     await utils.setLang('en');
+    await browser.sleep(100);
     expect(element(by.css('.alert.alert-info')).element(by.css('a')).getText()).toContain('undelegated');
     await utils.setLang('fr');
+    await browser.sleep(100);
     expect(element(by.css('.alert.alert-info')).element(by.css('a')).getText()).toContain('non délégué');
     await utils.setLang('sv');
+    await browser.sleep(100);
     expect(element(by.css('.alert.alert-info')).element(by.css('a')).getText()).toContain('odelegerat domäntest');
   });
 });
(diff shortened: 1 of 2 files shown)


matsduf (approved):
I think this looks fine.

blacksponge:
After a more thorough check, is seems that the loading of this json was causing delays so I added a sleep. It would have been better to use `browser.wait` but I could figure what a reliable wait condition could be.

e2e/FR16.e2e-spec.ts
```diff
@@ -16,12 +16,15 @@ describe('Zonemaster test FR16 - [The advanced view should have a text describin
     expect(element(by.css('.alert.alert-info')).element(by.css('a')).getAttribute('fragment')).toBe('q12');
   });
 
-  it('should have a description text in multi languages', () => {
-    utils.setLang('en');
+  it('should have a description text in multi languages', async () => {
+    await utils.setLang('en');
+    await browser.sleep(100);
```
matsduf, line 21: Is that seconds? Isn't 100 s beyond what is needed?

blacksponge: It is in milliseconds

matsduf: > It is in milliseconds

Then it is more than fine. :-)

matsduf (approved):

mattias-p:
I'm going way out on a limb here but after #222, couldn't you do something like this? (Keep in mind that I have neither tried this myself, nor do I really know any of this stuff.)

```typescript
browser.wait(protractor.ExpectedConditions.presenceOf($('.lang > div > a.selected[lang="' + lang + '"]')), 5000);
```

Also, wouldn't it make sense to move the waiting into util.setLang()?

blacksponge:
It should be possible to do something with `browser.wait`, I'll have another look into that.

blacksponge:
The changes now live in #222, as they depend on the changes there.

matsduf:
> The changes now live in #222, as they depend on the changes there. If that's okay we can close this PR.

Please update the description of #222 and refer to this PR.

blacksponge:
> Please update the description of #222 and refer to this PR.

Done.

Issue #178, opened 2021-02-04
Highlight invalid fields


mattias-p:
Once zonemaster/zonemaster-backend/issues/703 is implemented it becomes possible for GUI to give user feedback on what fields contain validation errors. The GUI should take this opportunity to highlight the offending fields in order to improve the user experience.

blacksponge:
resolved by #230

Issue #177, opened 2021-02-04
Better error message for validation errors


mattias-p:
When you press "Check" to check a domain name, a start_domain_test request is sent to the Backend. If an error response is returned, the message "Internal server error" is displayed. If the error response is due to invalid input submitted by the user, the displayed message is still "Internal server error". This is not helpful to the user.

It would be better if the GUI inspected the error code and displayed a different message if the error is caused by invalid inputs.

blacksponge:
resolved by #230

Issue #175, opened 2021-02-04
Sanity checks on nameservers


matsduf:
If a name server name is empty but an IP is specified, then that name server is ignored and the test is started without that information. The user should be given information that name server name is required.

If the name server name is " " then that name is accepted and sent to the server. The GUI should remove leading and trailing spaces.

blacksponge:
resolved by #230

PR #230, opened 2021-06-17
Improve errors

blacksponge:
## Purpose

Display better error message instead of just "Internal server error".

- [x] All field have a associated feedback block.
- [x] Move the feedback below the associated field (for ns/ds field groups).
- [x] Improve readability (currently the font size and colour make reading difficult).

## Context

Addresses #175, #177 and #178, depends on <https://github.com/zonemaster/zonemaster-backend/pull/820> 

## Changes

In order to easily map the error message to the fields, some refactor has been done to use only one FormGroup.

The FormGroup being a mapping to the request param allows to the field using the json path returned by the server. The error is then injected into the form.

~**BREAKING**: Because of the new language parameter included in the requests for error translations, this PR requires the backend to have a version >= v2021.2.~ Will be added in a subsequent PR.

## How to test this PR

Fill garbage in the form and see if appropriate errors are displayed.


blacksponge:
WIP:
Screenshot 2021-06-17 at 16-26-46 Zonemaster

matsduf:
I think this looks like a good attempt to get better messages to the user. An improvement to can give much better user experience.

The screen shots shows that we need more user friendly error messages. We could have detailed information on what is expected in each field via a link.

This depends on zonemaster/zonemaster-backend#795, and I suggest that we try to have to set before the work on this has gone too far.

mattias-p:
I really like what I see in the screenshots!

matsduf:
> I really like what I see in the screenshots!

That could work for CLI, but not for GUI.

mattias-p:
> That could work for CLI, but not for GUI.

Can you elaborate?

matsduf:
@blacksponge, is this ready for review?

blacksponge:
It is, yes, as is zonemaster/zonemaster-backend#795 by the way.

matsduf:
> It is, yes, as is [zonemaster/zonemaster-backend#795](https://github.com/zonemaster/zonemaster-backend/pull/795) by the way.

I will review both.

matsduf:
> It is, yes, as is [zonemaster/zonemaster-backend#795](https://github.com/zonemaster/zonemaster-backend/pull/795) by the way.

Does this depend on zonemaster/zonemaster-backend#795? If so, please update the description with that.

matsduf:
This does not work with un-modified Backend. Request:
```json
{"jsonrpc":"2.0","id":1625146491663,"method":"start_domain_test","params":{"language":"en","domain":"iis.se","ipv4":true,"ipv6":true,"profile":"default","nameservers":[],"ds_info":[],"client_version":"3.3.0","client_id":"Zonemaster GUI"}}
```
Response:
```json
{"error":{"code":"-32602","message":"Invalid method parameter(s).","data":"/: Properties not allowed: language."},"id":1625146491663,"jsonrpc":"2.0"}
```

Please update the description to state the requirements.

matsduf (changes requested):
This does not work with un-modified Backend. Request:
```json
{"jsonrpc":"2.0","id":1625146491663,"method":"start_domain_test","params":{"language":"en","domain":"iis.se","ipv4":true,"ipv6":true,"profile":"default","nameservers":[],"ds_info":[],"client_version":"3.3.0","client_id":"Zonemaster GUI"}}
```
Response:
```json
{"error":{"code":"-32602","message":"Invalid method parameter(s).","data":"/: Properties not allowed: language."},"id":1625146491663,"jsonrpc":"2.0"}
```

Please update the description to state the requirements.

blacksponge:
> Does this depend on zonemaster/zonemaster-backend#795? If so, please update the description with that.

It is already stated on the description.

> Please update the description to state the requirements.

I have updated the description, is that better?

matsduf:
> > Does this depend on [zonemaster/zonemaster-backend#795](https://github.com/zonemaster/zonemaster-backend/pull/795)? If so, please update the description with that.
> 
> It is already stated on the description.

I could not see that.

> > Please update the description to state the requirements.
> 
> I have updated the description, is that better?

Yes, but rather refer to that it depends on the update in zonemaster/zonemaster-backend#795, and that it will not work on an unmodified Backend version.

matsduf:
Cannot install the dependency zonemaster/zonemaster-backend#795. Waiting for that to be fixed.

matsduf:
This seems to work with zonemaster/zonemaster-backend#795, but I found an error there. See the other PR.

One comment: If I get an error and switch language, e.g. from French to English, the everything but the error message is presented in the new language. How difficult would it be to make the message change language?

blacksponge:
> If I get an error and switch language, e.g. from French to English, the everything but the error message is presented in the new language. How difficult would it be to make the message change language?

As discussed previously here https://github.com/zonemaster/zonemaster-backend/pull/795#discussion_r657870475 the current implementation does not allow for easy "on the fly" translation of the backend served error message and that would means re-submitting the form which I don't think is a good idea. To be able translate the messages without re-submitting the form we need to serve separate translation files, ex. json files served by the backend which would need an updated translation logic on the client side to support multiple translation source.

matsduf:
> As discussed previously here [zonemaster/zonemaster-backend#795 (comment)](https://github.com/zonemaster/zonemaster-backend/pull/795#discussion_r657870475) the current implementation does not allow for easy "on the fly" translation of the backend served error message, that would mean re-submitting the form which I don't think is a good idea.

I think the limitation could be acceptable, but I do not see the problem of re-transmitting the call. Nothing has been added to the database and this is probably not heavier for RPCAPI than a request of the result in another language. What is the
problem of re-transmitting the call?

If we stay with how it is (no language update) then I would like to see it documented in GUI as a known limitation. Maybe in the README file.

blacksponge:
> What is the problem of re-transmitting the call?

I think of that more in a UX point of view, if someone change the language while in the middle of editing the form, the domain test could start before the user has finished filling it form. 

What we could do though, is adding a "dry run" option that only check the parameters without actually starting the test. That could be fairly easy to implement, but maybe as an subsequent PR.

matsduf:
> I think of that more in a UX point of view, if someone changes the language while in the middle of editing the form, the domain test could start before the user has finished filling it form.

When I tested the messages came after pressing the "check" button. If you change language before that it should not change anything. If press "check" and get error messages and change language, a re-transmission should get the same error messages, but in the new language.

blacksponge:
> When I tested the messages came after pressing the "check" button. If you change language before that it should not change anything. If press "check" and get error messages and change language, a re-transmission should get the same error messages, but in the new language.

I will see what I can do for server validation error translation in the GUI, but probably not in this PR.

blacksponge:
>  but probably not in this PR.

Opening https://github.com/zonemaster/zonemaster-gui/issues/239 to track that.

matsduf:
This PR looks fine, but I want to do a final check against a corrected  zonemaster/zonemaster-backend#795 first.

matsduf (dismissed):

matsduf:
@mattias-p and @pnax, can you review this?

matsduf (dismissed):

src/assets/i18n/fr.json
```diff
@@ -29,6 +29,7 @@
 	"Message": "Message",
 	"Module": "Module",
 	"Nameservers": "Serveurs DNS",
+	"NS required": "NS obligatoire",
```
pnax, line 32: There are some other new translatable strings (e.g. 'Digest type required', 'Algorithm required', ...). Do you plan on providing a translation within this PR? Or should we do it in other PRs?

pnax (dismissed):
Looks pretty nice. This is a great step forward!

src/assets/i18n/fr.json
blacksponge: I've updated the translation.

src/app/components/form/form.component.html
```diff
@@ -79,13 +79,13 @@ <h1>{{'Domain name'|translate}}</h1>
                 <div class="custom-control custom-checkbox custom-control-inline">
                   <input type="checkbox" class="custom-control-input" formControlName="disable_ipv4" id="disable_protocol_ipv4" [class.is-invalid]="newForm.errors?.noProtocol || newForm.controls.disable_ipv4.invalid">
                   <label class="custom-control-label" for="disable_protocol_ipv4">
-                    {{'Disable IPv4' |translate}}
+                    {{ 'DISABLE_IP_CHECKBOX' | translate: {ip: 'IPv4'} }}
```
pnax, line 82: Since you have explicitly defined the string, I think you can keep the full string instead of using the variable I introduced in https://github.com/zonemaster/zonemaster-gui/pull/243/commits/710f06ef2cb7ac501b6564d9ed6cf25f828394af.
(it would imply that each LANG.json should be updated with the full string as added in #218)

src/app/components/form/form.component.html
```diff
@@ -79,13 +79,13 @@ <h1>{{'Domain name'|translate}}</h1>
...
                   </label>
                 </div>
                 <div class="custom-control custom-checkbox custom-control-inline">
                   <input type="checkbox" class="custom-control-input" formControlName="disable_ipv6" id="disable_protocol_ipv6" [class.is-invalid]="newForm.errors?.noProtocol || newForm.controls.disable_ipv6.invalid">
                   <label class="custom-control-label" for="disable_protocol_ipv6">
-                    {{'Disable IPv6' |translate}}
+                    {{ 'DISABLE_IP_CHECKBOX' | translate: {ip: 'IPv6'} }}
```
pnax, line 88: same

pnax (commented):
Just a remark on the merging of develop into this branch

matsduf (dismissed):

matsduf:
I suggest that we have this merged and if further refinements are needed, they can be handled in a new PRs. This PR is quite large now. @pnax, can you approve the PR with the new changes?

pnax (dismissed):
Just a question, should we wait for https://github.com/zonemaster/zonemaster-backend/pull/795 to be merged first?

blacksponge:
> Just a question, should we wait for zonemaster/zonemaster-backend#795 to be merged first?

Yes, it's better

matsduf:
> Just a question, should we wait for [zonemaster/zonemaster-backend#795](https://github.com/zonemaster/zonemaster-backend/pull/795) to be merged first?

In principle, yes, but I do not think it matters. @pnax, please review zonemaster/zonemaster-backend#795.

blacksponge:
I disabled the language feature (for error translation) the time the changes in backend make it to develop, this way the GUI is still compatible with the current error handling without breaking.

@matsduf can you re-approved so I can at least merge this?

blacksponge:
Actually I might have to change some stuff in the tests also...

matsduf:
> Actually I might have to change some stuff in the tests also...

I will test when it passes Travis.

matsduf:
The images are no longer relevant. Almost all input errors result in "internal server error". Empty input domain and some errors in the permitted characters in the input domain are captured in the GUI, but not the most obvious ones, i.e. leading an trailing space.

matsduf (approved):

matsduf:
See #249.

matsduf:
When testing against zonemaster/zonemaster-backend#820 (https://github.com/zonemaster/zonemaster-backend/pull/820/commits/589fa8b4962171bffcb374c32e80480b47f5155a) errors are displayed as shown above. If space is included a direct error message is shown.

blacksponge:
Can this be merged?

matsduf:
> Can this be merged?

Yes, I think so. We have tested it, but now it is better if we use it and go forward. Before release we will test it again. If you do not hesitate or @pnax does not object, go ahead today.

pnax:
## Testing v2021.2

The errors are properly displayed. However, errors are not fully translated for Danish, Finnish, Norwegian and Swedish. Because this PR is not about translating the error, tlgtm.

matsduf:
> The errors are properly displayed. However, errors are not fully translated for Danish, Finnish, Norwegian and Swedish. Because this PR is not about translating the error messages, tlgtm.

@pnax, have you really used the latest update of the Zonemaster-Backend develop branch? There is since some time an updated translation to Swedish, [sv.po](https://github.com/zonemaster/zonemaster-backend/blob/develop/share/sv.po). Or have you found a bug in the code?

pnax:
My remark was not about Backend error messages, but GUI error messages displayed under each input field.

matsduf:
I guess some come from Backend and others come from [sv.json](https://github.com/zonemaster/zonemaster-gui/blob/develop/src/assets/i18n/sv.json) (in case of Swedish). There are three untranslated strings for Swedish that I have missed:
```json
    "Algorithm required": null,
    "Digest type required": null,
    "NS required": null,
```
Do you mean those?

pnax:
Well it seems I got things wrong here. At first the message errors are the one from the GUI, then when changing language, the form is submitted and then we get the error messages from the Backend and not from the GUI anymore. Using the latest develop from Backend, I now have proper translations for Swedish. My bad. And indeed the 3 messages that you listed are missing in the GUI. So they are not translated when filling the form the first time.

matsduf:
> And indeed the 3 messages that you listed are missing in the GUI. So they are not translated when filling the form the first time.

An update is provided in #286.

PR #234, opened 2021-06-30
don't collapse expended modules when language changes

blacksponge:
## Purpose

When someone changes the language once the results are displayed, if one of the module is expended then is will be collapsed.

## Changes

This PR changes the behaviour of the result component to not re-collapse modules if the language has changed. 

When the result id changes the behaviour remains the same as before.

## How to test this PR

* Open a test results (ex. http://localhost:4200/result/df0aa98ff4e53b99 if your dev environment point to zonemaster.net)
* Expend a module
* Change the language
* If the module remains expended, it is working as expected
* Now try to change the result id by displaying another test from the history
* If all modules are now collapsed, it is working as expected

### DIFF
--- a/e2e/FR22.e2e-spec.ts
+++ b/e2e/FR22.e2e-spec.ts
@@ -18,11 +18,11 @@ describe('Zonemaster test FR22 - [Provide the possibility to see more informatio
     await element(by.css('#domain_check_name')).sendKeys('afNiC.Fr');
     await element(by.css('div button.launch')).click();
 
-    await browser.wait(() => element(by.css('#collapsed_module_1')).isPresent(), 120 * 1000);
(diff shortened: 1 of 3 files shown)


matsduf (approved):
This looks fine. It works as it should when tested. It improves the appearance when changing language.

Also here I would like to see more documentation of the code. In Perl there is POD for documentation. Is there something similar for this code.

pnax (approved):
Nice feature.

matsduf:
@blacksponge, please merge.

pnax:
## Testing v2021.2

- [x] If the module remains expended, it is working as expected
- [x] If all modules are now collapsed, it is working as expected

LGTM

Issue #229, opened 2021-06-16
Remember the selected language


blacksponge:
Currently if I select a language and reload the page, it is back to the default language. Same thing if I close the tab and come back later. It would be better if the language was stored (for instance in local storage) and restored in subsequent visit.

matsduf:
When the GUI is loaded it will look for your default language setting. So if that is fr_FR then you will get i in French. If the language does not match any of the Zonemaster languages, then you will get i in the default language (English). That should work for most users, I guess. I think we should avoid setting any cookies.

blacksponge:
> That should work for most users, I guess.

Since we are offering a language selection menu, I think that for it to work correctly the user should not have to change their language each time they open the website.

> I think we should avoid setting any cookies.

Yes that's why I am talking about local storage, we don't need to send that info to the backend.

blacksponge:
resolved by #235

PR #235, opened 2021-06-30
store language in local storage

blacksponge:
## Purpose

Remember the language so someone who change it does not have to set it each it they use zonemaster.

## Context

Addresses #229

## Changes

Stores the language in local storage and retrieves it at load time. If no language is found is storage defaults on the browser language.

## How to test this PR

* Change the language
* Reload
* Check that the selected language is the same

### DIFF
--- a/src/app/components/navigation/navigation.component.ts
+++ b/src/app/components/navigation/navigation.component.ts
@@ -17,7 +17,7 @@ export class NavigationComponent implements OnInit {
 
   constructor(private translateService: TranslateService, zone: NgZone) {
     this.translateService.setDefaultLang(this.lang_default);
-    this.lang = this.translateService.getBrowserLang();
+    this.lang = localStorage.getItem('lang') || this.translateService.getBrowserLang();
     if (this.isValidLanguage(this.lang)) {
       this.setLanguage(this.lang);
     } else {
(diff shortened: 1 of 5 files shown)


src/app/components/navigation/navigation.component.ts
```diff
@@ -17,7 +17,7 @@ export class NavigationComponent implements OnInit {
 
   constructor(private translateService: TranslateService, zone: NgZone) {
     this.translateService.setDefaultLang(this.lang_default);
-    this.lang = this.translateService.getBrowserLang();
+    this.lang = localStorage.getItem('lang') || this.translateService.getBrowserLang();
```
matsduf, line 20: Where is this local storage? Is the local storage bound to where the URL where the GUI was loaded from?

blacksponge: It is bound the origin of the web page so essentially the protocol, domain and port used to load it.

https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage

matsduf: And that is a standard so that it works on any browser, or at least does not create problem on other browsers?

blacksponge: It is a standard, and as you could see on the page linked above, it is very well supported.

https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage#browser_compatibility

matsduf (approved):
It is a nice feature. I works as it should.

When I look at the code I see no documentation. Could you add some documentation to the additions and updates that you do? To make it easier for others to update in the future.

pnax (approved):

matsduf:
@blacksponge, please merge.

mattias-p:

Issue #153, opened 2020-10-29
Draw up a color scheme


mattias-p:
Today the use of colors in the UI is quite undisciplined. For example, I can find four visual elements on the results page that use some tone of blue. Among them they use three different tones of blue, and the tones don't particularly fit together. Moreover, as pointed out in #150, the blue background color in the test results make them hard to read. And that's just the blues. I have complaints for the other colors too.

We should have a proper color scheme drawn up and documented for Zonemaster GUI. Then we should stick to that color scheme for all visual elements in the UI.

pnax:
From https://github.com/zonemaster/zonemaster-gui/issues/150#issuecomment-779855697, and the link to the Excedo GUI, I have made some screenshots to compare both GUIs. I am not an UX/UI designer but I feel that Excedo's GUI is sober and hence more accessible.

I think that :

1. we should remove the background color in the array and use a similar approach than Excedo and use colors as an addition to the text
2. as stated in this issue, we should use less colors, or stick the colors to the results, so I would suggest to use the same color for all buttons `History`, `Export`, `Share` and `Settings` (as done by Excedo)
3. we should rethink which color we use for what and document that
4. we should use better contrast when the text is written with a background color


@matsduf @mattias-p any thoughts on this ?


## Screenshots

### The result array

Zonemaster
![zm-results](https://user-images.githubusercontent.com/70589067/112961841-07d3ee80-9146-11eb-9f69-2c17caa4a21c.png)
Excedo
![excedo-results](https://user-images.githubusercontent.com/70589067/112961828-060a2b00-9146-11eb-8ddd-7dead10d3cdf.png)


### The filter buttons

Zonemaster
![zm-filters](https://user-images.githubusercontent.com/70589067/112961834-073b5800-9146-11eb-8648-6a7feebfaa60.png)
Excedo
![excedo-filters](https://user-images.githubusercontent.com/70589067/112961825-04d8fe00-9146-11eb-9e35-9ac8b09a7d51.png)

### INFO:

Zonemaster
![zm-info](https://user-images.githubusercontent.com/70589067/112961836-073b5800-9146-11eb-9d41-d09a57db3c33.png)
Excedo
![excedo-info](https://user-images.githubusercontent.com/70589067/112961826-05719480-9146-11eb-9506-c6b4d273dad6.png)

### NOTICE:

Zonemaster
[...]

Issue #150, opened 2020-10-29
UX/UI Color contrast for results table


pnax:
In the result table, the current blue background color `#007bff` makes it hard to read the content of the cell written in black.

Is it possible to use better contrast between text color and background color ?

matsduf:
Excedo has installed Zonemaster with some customization. Take a look at that. https://zonemaster.excedonet.net/domain_check

pnax:
Resolved with #236

PR #236, opened 2021-07-01
[proposal] Removing background colors in results array and buttons

pnax:
## Purpose

This is a suggestion to have less colors on the UI. This is also about sharing the code I wrote for https://github.com/zonemaster/zonemaster-gui/issues/153#issuecomment-811107482

## Context

Try to address #150, #153

## Changes

* Removes the line number in the result array

* Set new column width

* Defines new colors

| Level | color |
| --- | --- |
| INFO | `#66ad54` |
| NOTICE | `#007bff` |
| WARNING | `#ff9019` |
| ERROR | `#ec3545` |
| CRITICAL | `#ec3545` |

### Collapsed table

zm-gui1

### Filters and buttons

zm-gui2

### Expand some modules

zm-gui3

### Do not group by modules

zm-gui4

## How to test this PR

Only affecting CSS, the testing is made visually.

### DIFF
--- a/src/app/components/result/result.component.css
+++ b/src/app/components/result/result.component.css
@@ -85,6 +85,19 @@
   transition: all 300ms ease-in-out;
 }
 
+
+.thead-dark tr th:first-child {
+  width: 40px;
+}
+
+.thead-dark tr th:nth-child(2) {
+  width: 15%;
+}
+
+.thead-dark tr th:nth-child(3) {
+  width: 12%;
+}
+
 .active.info, .result.INFO {
   background-color: #007bff;
 }
--- a/src/app/components/result/result.component.css
+++ b/src/app/components/result/result.component.css
@@ -129,8 +129,16 @@ tbody[id^=collapsed_module] tr:first-child th::before {
     position: absolute;
     width: 7px;
     height: 100%;
-    right: 0;
+    left: 0;
     top: 0;
+}
+.result.collapse.INFO:not(.expand) th:first-of-type::after,
+.result.collapse.NOTICE:not(.expand) th:first-of-type::after,
+.result.collapse.WARNING:not(.expand) th:first-of-type::after,
+.result.collapse.ERROR:not(.expand) th:first-of-type::after,
+.result.collapse.CRITICAL:not(.expand) th:first-of-type::after {
+    right: 0;
+    left: auto;
 }
 
 .active.info,
--- a/src/app/components/result/result.component.css
+++ b/src/app/components/result/result.component.css
@@ -99,19 +99,19 @@
 }
 
 .active.info {
-  background-color: #007bff;
+  background-color: #66ad54;
 }
 .active.notice {
-  background-color: #fff638;
+  background-color: #007bff;
 }
 .active.warning {
   background-color: #ff9019;
 }
 .active.error {
-  background-color: #dc3545;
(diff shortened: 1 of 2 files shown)


matsduf:
It looks good. I will run a test soon.

matsduf:
Should the colors for ERROR and CRITICAL really be identical? I think it would be good if we could involve someone working with web and graphical interfaces at Afnic or IIS to suggest what colors we should use. I think info and notice could be less intensive.

pnax:
> Should the colors for ERROR and CRITICAL really be identical?

No idea. To be honest I don't know what are the standards out there.

> I think it would be good if we could involve someone working with web and graphical interfaces at Afnic or IIS to suggest what colors we should use.

I agree. I see if I can have some reviews.

> I think info and notice could be less intensive.

Do you mean a less intensive color ?

matsduf:
> > I think info and notice could be less intensive.
> 
> Do you mean a less intensive color ?

Yes, I think they stick out too much.

blacksponge:
This is a really nice improvement on readability and overall consistency!

blacksponge:
The "All" button has the same colour as "Notice", maybe it should be gray instead, the same way the "History" and other buttons are?

matsduf:
This also raises that question on what level that messages should be shown from by default. The CLI starts with NOTICE by default. To get more information you could select to show from INFO (`--level INFO`). Here in GUI all messages available through the API are shown, and that starts with INFO.
1. There is no possibility to get more information since all information is already there.
2. To avoid overloading the user the amount of INFO messages must be kept down. There are already too many of them if the INFO level is the default level.

I think that the presentation would probably be better if it by default present NOTICE and higher, and makes it possible for the user to select "more information", maybe on module level or even on test case level (requires more development).

Is this relevant for this PR? Both yes and no. This PR can probably be merged with some color adjustments, but it could also trigger some more improvements in other PRs.

matsduf (approved):
I think the "History", "Export" and "Share" buttons could keep their colors, but overall I think this is a large improvement. I think we can go ahead with this as-is, and then try to find someone with a trained eye for user interface who could suggest improvements.

benoitam (commented):
Very nice improvement. I agree with the suggested changes.

blacksponge (dismissed):

matsduf:
@pnax, please resolve the conflict and I will approve again.

pnax:
@matsduf 
> > Do you mean a less intensive color ?
>
> Yes, I think they stick out too much.

I made some tries, but I couldn't get something satisfactory.

@blacksponge 
> The "All" button has the same colour as "Notice", maybe it should be gray instead, the same way the "History" and other buttons are?

Using the same value used for the other buttons is in conflict with the background of the badge holding the count of items. However I found that using `#919191` might be ok. I made an update with that value.

matsduf (approved):
I approve this because I think this is a great improvement, but I still think that the "History", "Share" etch buttons could have some color. Now it is too grey. It could be adjusted in a future PR and I still think we should engage somebody who is trained in constructing graphical interface.

image

pnax:
> I still think we should engage somebody who is trained in constructing graphical interface.

I totally agree on this.

pnax:
## Testing v2021.2

Some PR have updated the output a bit. However the overall result is still compliant with the idea behind this PR. LGTM.

PR #237, opened 2021-07-01
Remove the test number in the results

pnax:
## Purpose

Currently the result page highlight the test number more than the domain name. This looks like an odd choice as we don't use this number anywhere. This PR removes the test number and highlights the domain instead.

This test id also leaks the number of tests in database.

## Context

n/a

## Changes

* Removes the test number
* Highlights the domain name

### Before

zm-gui-header-1

### After

zm-gui-header-2

## How to test this PR

Verify that only the domain name is displayed.

### DIFF
--- a/src/app/components/result/result.component.html
+++ b/src/app/components/result/result.component.html
@@ -4,7 +4,7 @@
 <div class="result container" #resultView>
   <div class="row">
     <div class="col-md-6">
-      <h2>{{'Test #'|translate}}{{ test.id }}<small> - {{form.domain}}</small></h2>
+      <h2>{{form.domain}}</h2>
       <i>{{ test.creation_time | amFromUtc | date:'yyyy-MM-dd HH:mm zzzz' }}</i>
     </div>
     <div class="col-md-6">


matsduf:
Or maybe using the hash-ID and putting that after the domain name:
```
afnic.fr (test # 11d9991640c383f6)
```

But I do agree that the test number sticks out too much.

blacksponge:
> Or maybe using the hash-ID and putting that after the domain name

What is the value of displaying the id / hash to the user?

matsduf:
> > Or maybe using the hash-ID and putting that after the domain name
> 
> What is the value of displaying the id / hash to the user?

If you get a screen shot you can find the same test.

matsduf (approved):
I think the result is fine. It makes it much easier to read the relevant information.

It is useful to have the test ID (or rather hash test ID) but that can be fetched via "share" (also see #238). The test time and date is there, which makes it possible to find the exact test result.

matsduf:
@blacksponge and @mattias-p, please review.

pnax:
> It is useful to have the test ID

I agree, I thought a little bit on having the hash displayed somewhere, and I think one of the best place might be in the URL directly. So that the URL can be copy/paste (this can be another improvement added with your suggestion to improve the "Share" UI/UX).

blacksponge (approved):
I don't see much use of the hash id in the title but it could be kept as a separate information like the date. If it is planned to improve the share UX to show the link directly that could be a way to do it.

matsduf:
@pnax, please merge.

pnax:
## Testing v2021.2

lgtm

Issue #160, opened 2020-12-16
Document NGX-Translate-extract


matsduf:
This issue replaces first half of #143.

[NGX-Translate-extract] has been intruduced in #121 for handling translation. The process for updating the keys in the "src/assets/i18n/xx.json" files should be documented including where to download the [NGX-Translate-extract] program. The documentation should either be in [ReleaseProcess-preparation] or in Zonemaster-GUI and linked from [ReleaseProcess-preparation] .

[ReleaseProcess-preparation]: https://github.com/zonemaster/zonemaster/blob/develop/docs/internal-documentation/maintenance/ReleaseProcess-preparation.md
[NGX-Translate-extract]: https://github.com/biesbjerg/ngx-translate-extract

matsduf:
It is better to include the step when new tags are added in GUI.

Issue #241, opened 2021-07-06
Missing translation on "bubbles"


matsduf:
When you put the mouse over the Info, Notice etc buttons a "bubble" with explanation comes up. That is always in English. See screen shot where language is in Swedish except for the text in the "bubble".

<img width="1123" alt="image" src="https://user-images.githubusercontent.com/5430733/124568538-52e85680-de45-11eb-9875-0dc3aac72865.png">

matsduf:
Will be resolved by #243.

PR #121, opened 2020-04-26
Improve i18n

pamasse:
Like PR #116, this PR try to improve I18n support by using some new tool to align the translation files ({en,fr,da,sv}.json) to the code.
In order to use ngx-extractor, we need to use correctly our translator tool. So, I update some piece of codes to correctly use the translator service (this.translateService.get()).

I add two commands to the project:
-  i18n:init : Create a template.json file with all keys used in the code 
- i18n:extract: Update {en,fr,da,sv}.json files with news keys if detected or delete used ones.

to use these command, simply run `npm run i18n:extract` for example.

PR #243, opened 2021-07-06
Make it possible to translate buttons' tooltips/bubbles

pnax:
## Purpose

The button tooltips are not translated. This makes it possible to use the translation file for each language.

## Context

Addresses #241 

#### Main change
* support for tooltip translation using the `ngx-translate` and `ngx-extract` tools
* translate into French, Swedish (so far)

#### Side changes
* update the way we translate the "Disable IPv4" and "Disable IPv6" texts
* update the npm `i18n:extract` command to use 4 spaces instead of tabulations
* update the npm command to use _null_ value for unset translations
* run the `npm i18n:extract` command: this has sortes alphabetically each LANG.json file and replaced the tabs with 4 spaces
* reduce the length of the _copyright_ key

## How to test this PR

Set the GUI into English, then hover on the "All", "Info", "Notice", "Warning", "Error" and "Critical" buttons, make sure that there is a text for each tooltip (this should be the same as the one before the PR).
(if translations have been added within this PR, change the GUI language and do the same, the tooltip's text should be updated accordingly)

Side test, make sure that the "Disable IPv4" and "Disable IPv6" are correctly translated.

#### TODO

- [ ] translate into Danish
- [ ] translate into Finnish
- [x] translate into French
- [ ] translate into Norwegian
- [x] translate into Swedish


matsduf:
I think there should be a tool to create en.json from the code, and maybe also update all other json files with missing keys.

matsduf:
In the comment https://github.com/zonemaster/zonemaster-gui/pull/121#issuecomment-699900918 @pamasse writes about a tool to create the json files.

pnax:
Indeed, there is the command `npm run i18n:extract`. Sadly this will use tabulations instead of spaces in th LANG.json files, but it looks like we can update the command to use spaces.
https://www.npmjs.com/package/ngx-translate-extract

matsduf:
For sv.json:
```
    "TOOLTIP_ALL": "Visa alla resultat",
    "TOOLTIP_CRITICAL": "Critical - betyder att det finns ett mycket allvarligt fel.",
    "TOOLTIP_ERROR": "Error - betyder att det finns ett fel som det är mycket troligt ger negativa konsekvenser för hur den testade zonen fungerar, men inte så allvarligt att hela zonen är ouppslagbar.",
    "TOOLTIP_INFO": "Info - något som kan vara av intresse för zonens administratör men som helt klart inte tyder på att vara ett problem.",
    "TOOLTIP_NOTICE": "Notice - betyder något som zonens administratör bör känna till men som inte behöver vara något problem.",
    "TOOLTIP_WARNING": "Warning - betyder något som under vissa omständigheter kan vara ett problem, men det är inte sannolikt att en vanlig användare upptäcker det.",
```

matsduf:
@repperille, @haarbo and @samisalmensuo, can you provide a translation for the following lines? You can provide the translation as a comment in this PR, but make sure to retain the format.
```
    "TOOLTIP_ALL": "Display all results",
    "TOOLTIP_CRITICAL": "Critical - means a very serious error.",
    "TOOLTIP_ERROR": "Error - means a problem that is very likely (or possibly certain) to negatively affect the function of the zone being tested, but not so severe that the entire zone becomes unresolvable",
    "TOOLTIP_INFO": "Info - is something that may be of interest to the zone's administrator but that definitely does not indicate a problem.",
    "TOOLTIP_NOTICE": "Notice - means something that should be known by the zone's administrator but that need not necessarily be a problem at all",
    "TOOLTIP_WARNING": "Warning - means something that will under some circumstances be a problem, but that is unlikely to be noticed by a casual user.",
```

matsduf:
> Sadly this will use tabulations instead of spaces in th LANG.json files, but it looks like we can update the command to use spaces.

Spaces are better, but it is probably better to use a tool if possible. And converting from TAB to SP*4 before `git add` is very simple.

pnax:
> Spaces are better, but it is probably better to use a tool if possible.

I discovered and updated the npm command to use 4 spaces (see 4dd30ef).

I also updated on key, the one with the copyright text so that it takes less space in the file.

pamasse (approved):

matsduf:
> > Spaces are better, but it is probably better to use a tool if possible.
> 
> I discovered and updated the npm command to use 4 spaces (see [4dd30ef](https://github.com/zonemaster/zonemaster-gui/commit/4dd30ef18177b3e6dbb5d7462ddc76353a4f9130)).

I think we should have this documented somewhere. Unless we have a good place, could you just describe it in a new document under "docs"?

> I also updated on key, the one with the copyright text so that it takes less space in the file.

There are pros and cons with having keys. I thought of replacing all messages with keys, but that will make translation more complex since you have to look into two files.

matsduf (changes requested):
This looks great except that languages without translation get the tag instead of the English text.

Finnish selected:

image

pamasse:
> This looks great except that languages without translation get the tag instead of the English text.
> 
> Finnish selected:
> 
> image

I'm not 100% sure, but it's not the "normal" behavior? It's why we use "English sentence" as tag

pnax:
> This looks great except that languages without translation get the tag instead of the English text.

This is how it should work when using tag (instead of explicit string for the tag/key). The `npm run i18n:extract` command is set with `--key-as-default-value` which means that the translation is automatically populated with the key. We can choose to set it to the empty string (or null) with `--null-as-default-value`. So when the key is the English translation, the value is then already in English.

I made a test. If you set a value to `null` (for instance `"TOOLTIP_INFO": null`), then the English translation will be used instead.

1. change `--key-as-default-value` into `--null-as-default-value`
2. open the file fi.json and remove the line with `TOOLTIP_INFO`
3. run `npm run i18n:extract`
4. run the GUI and look for the tooltip text for the "INFO" button (it should be in English)

This might do the trick. However, I think it will also populate the en.json file with `null` values for new entries. I don't know what is best yet.

matsduf:
> This is how it should work when using tag (instead of explicit string for the tag/key). The `npm run i18n:extract` command is set with `--key-as-default-value` which means that the translation is automatically populated with the key. We can choose to set it to the empty string (or null) with `--null-as-default-value`. So when the key is the English translation, the value is then already in English.

I assume that `npm run i18n:extract` is run at develop time and the result is added to a PR, but of course that requires that it is documented somewhere.

> I made a test. If you set a value to `null` (for instance `"TOOLTIP_INFO": null`), then the English translation will be used instead.

That sounds like what we want. Could you replace with `null` in this PR?

> This might do the trick. However, I think it will also populate the en.json file with `null` values for new entries. I don't know what is best yet.

That should not be a problem because we have to add the English text as soon as we add a tag.

matsduf (dismissed):
This looks fine now. I will approve again when the conflicts are resolved.

pnax:
To resolve the conflicts I had to update the way the "Disable IPv[46]" translations are done, because otherwise they won't be caught by the `npm run i18n:extract` command (see 710f06e).

docs/TranslationGuide.md
```diff
@@ -25,6 +25,23 @@ following languages with the attached language code:
...
 
+## Extracting translatable strings
+
+When adding new translatable strings to the GUI, they need to be added to each
+*LANG.json* file. This can be done with the following command:
+
+```
+npm run i18n:extract
+```
+
```
matsduf, line 36: Thanks for the documentation. This step should be executed every time a new string is added or changed (in the code).

Furtherdown in this file there is a quote from `package.json`. Should the code in `package.json` be updated? And then also the quote?

matsduf (dismissed):
This looks good.

matsduf:
@blacksponge, @mattias-p and @pamasse, please re-review.

blacksponge (dismissed):
lgtm

docs/TranslationGuide.md
pnax: Good point.

pnax: > Should the code in package.json be updated?

I'm not sure what you are referring to here, the code has already been updated
https://github.com/zonemaster/zonemaster-gui/blob/fe29580429de92215006cea55087f641d6399711/package.json#L14

> And then also the quote?

I updated the documentation.

pnax: I found issue #160, do you think we need update the ReleaseProcess-preparation.md document as well?

matsduf: I missed the update of package.json, sorry.

It is probably good that is also done at package build, but then the change will not go back into the repository. A good thing about "null" is that it is visible.

To make this work I think we should have steps like the following:
1. A PR is created
2. A string to be translated is added or modified (could in clear text or just a tag).
3. The command is run to update *.json files
4. en.json is update (if it is a tag)
5. Possibly fr.json and sv.json are updated, but could be postponed
6. The PR is merged
7. Issues for remaining languages are created and those updates are handled in separate PRs.

No, I do not think we should handle that as part of the release process. I will re-review and possibly close the issue. The question is if we need more documentation here in GUI.

matsduf (approved):

docs/TranslationGuide.md
matsduf: I closed #160 with a comment.

blacksponge (approved):

matsduf:
@pnax, please merge.

mattias-p:
I can verify that these strings are properly translated into Swedish. Norwegian and Finnish don't have translations for these strings and they show up as English, which is also correct.
However there are both Danish and French translations for these strings, but these strings show up in English for those languages. I haven't figured out why this happens. Other strings are property translated to Danish and French.

matsduf:
@mattias-p, I suggest that you create an issue with with Bug label and v2021.2 milestone so that we can easily find it.

pnax:
@mattias-p I've just installed the GUI (867f12c) and I can't reproduce. The strings are properly translated into Danish, English, French and Swedish (Finnish and Norwegian translations are missing indeed).

mattias-p:
After reinstalling and clearing my browser cache I am unable to reproduce the above problem. However this begs another question. Will other browsers keep using old cached files after the web GUI they're using is updated?

Apart from this everything seems to work fine.

matsduf:
@mattias-p, are you ready to set the label "ReleaseTested" or should you create an issue for further investigations?

mattias-p:
@matsduf I created #283.

Issue #238, opened 2021-07-02
Make it easy to copy the link to the test result


matsduf:
When you click the "share" button, you get a small sub-window with a link to the test result, but the there is no easy way to copy that link. The sub-window is also placed far from the button.

* Put the sub-window just above the button.
* Include a "copy-link" button or something in the sub-window.

<img width="693" alt="image" src="https://user-images.githubusercontent.com/5430733/124266107-d39e0e80-db36-11eb-9aa4-61a3b2a8e576.png">

pnax:
Done with #247.

PR #247, opened 2021-07-06
Avoid using modal window

pnax:
## Purpose

This PR is a proposal to avoid using the modal window for the "Share" and "Export" buttons.
This also adds an icon to copy the link to the clipboard.

## Context

Addresses #238

### Share box
zm-gui-drop2

Once you click, if the URL has been copied, you get the following visual feedback:
zm-gui-drop3

this one otherwise:
zm-gui-drop4

### Export box

zm-gui-drop1

### Settings (right alignment as well)

zm-gui-drop5

## How to test this PR

1. copy/paste a garbage text
2. click on the "Share" button
2. click on the "copy" icon
3. you should get a green check icon
4. paste somewhere what is in the buffer
5. you should get the link (and not the garbage text)

### DIFF
--- a/e2e/FR25.e2e-spec.ts
+++ b/e2e/FR25.e2e-spec.ts
@@ -13,21 +13,21 @@ describe('Zonemaster test FR25 - [Should be able to export the result in multipl
   });
 
   it('should have an export button',  async() => {
-    await browser.wait(() => element(by.css('a.btn.export')).isPresent(), 120 * 1000);
+    await browser.wait(() => element(by.css('button.btn.export')).isPresent(), 120 * 1000);
 
-    await expect(element(by.css('a.btn.export')).getText()).toEqual('Export');
+    await expect(element(by.css('button.btn.export')).getText()).toEqual('Export');
   });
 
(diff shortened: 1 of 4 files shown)


blacksponge:
Would it be possible to align the dropdowns to the bottom right (`placement="bottom-right"`, see screenshot)? it is easier to then click on the copy button if it is below the other one.
Screenshot 2021-07-08 at 11-10-25 Zonemaster

pnax:
Sure, I like the idea. I will also add this alignment rule for the other ones to make it consistent.

matsduf (commented):
I think it would be better to have to pop-ups above the buttons. Else it looks OK besides that the settings pop-up is different.

blacksponge (approved):
works as expected, I think it is a nice improvement

pnax:
> I think it would be better to have to pop-ups above the buttons.

I tested it, and I am not convinced. Here are some screenshots (I didn't push the code to the branch, to test you need to change `placement="bottom-right"` into `placement="top-right"`)


gui-no-mod1
gui-no-mod2
gui-no-mod3

pnax:
@matsduf any remarks on the latest screenshots?

pnax:
## Testing v2021.2

lgtm

PR #248, opened 2021-07-07
Update Danish translation for tool tips

### LINKED ISSUE #244: Translation of buttons' tooltips/bubbles (Danish)
PR #243 provides a mechanism for translation of the severity level buttons' tooltips/bubbles. Can you provide a translation for these lines? Either as a PR for da.json (when #243 is merged) or as a comment to this issue, but make sure to retain the format.
```
    "TOOLTIP_ALL": "Display all results",
    "TOOLTIP_CRITICAL": "Critical - means a very serious error.",
    "TOOLTIP_ERROR": "Error - means a problem that is very likely (or possibly certain) to negatively affect the function of the zone being tested, but not so severe that the entire zone becomes unresolvable",
    "TOOLTIP_INFO": "Info - is something that may be of interest to the zone's administrator but that definitely does not indicate a problem.",
    "TOOLTIP_NOTICE": "Notice - means something that should be known by the zone's administrator but that need not necessarily be a problem at all",
    "TOOLTIP_WARNING": "Warning - means something that will under some circumstances be a problem, but that is unlikely to be noticed by a casual user.",
```

matsduf:
## Purpose

PR #243 adds mechanism for translation of tool tips. 

## Context

Resolves #244 with data provided by @haarbo in #244.

## Changes

Adds Danish translation to the tool tips in da.json.

## How to test this PR

Open Zonemaster GUI with this PR included, test some domain, switch to Danish translation and move the mouse over the "Info", "Warning" etch buttons and verify that the tool tips is in Danish.


pnax:
I haven't translated "Critical", "Error", ... into French, and I think you haven't either for Swedish. Should we translate each term (as the text in each button which is not currently translated) ?
It looks like it's related to #86.

matsduf:
> I haven't translated "Critical", "Error", ... into French, and I think you haven't either for Swedish. Should we translate each term (as the text in each button which is not currently translated) ?

I have thought about it. I kept the same term in the tool tip as on the button to have the clear connection. If the buttons are translated, to be consistent, we would then have to translate the same term in the report, don't we? Else it would be strange. With the translated tool tips you can get an explanation.

I think we can wait and see. We still have some time to go before next release.

matsduf:
I notice that @haarbo chose to translate the term in the tool tip for Danish. I guess that the natural operation for translation.

pnax:
> I guess that the natural operation for translation.

Yes I guess it too.

> Should I update to make it consistent with French and Swedish?

I would say yes, so that it is also aligned with the text in each button (even though this is a special case in the translation).

matsduf:
> > Should I update to make it consistent with French and Swedish?
> 
> I would say yes, so that it is also aligned with the text in each button (even though this is a special case in the translation).

I have done so. Please re-review, @pnax.

pnax (approved):

mattias-p:
I tested this for v2021.2 and it looks like Danish to me.

PR #256, opened 2021-08-04
improve development environment

blacksponge:
## Purpose

Make GUI development smoother (me getting tired of manually switching between dev and testing env).

## Changes

* Use a separate configuration for development and testing.
* Use local zonemaster in dev.
* Add proxy configuration to use the local zonemaster instance (testing against zonemaster.net is still possible by launching  `npm run ng serve -- --configuration=tests` ).

## How to test this PR

* Run `npm start` to serve the development server
* Run the API on the same machine
* Access your browser and check that you are actually using the local API
* Run the e2e tests, which use zonemaster.net, if they pass all is good


pnax (approved):
Works as expected. This is a nice improvement to our dev env indeed.

PR #257, opened 2021-08-04
change url when test is done

blacksponge:
## Purpose

Allows for easy copy/paste from the url bar and also push the url in the browser history.

## Changes

* Change the url to the test result when the test is done. 
* This **does not** route the user to the result screen, this way it is still possible to make another test without going back.

## How to test this PR

* Make a test
* Check that the url has change to the result url
* Test that it is still possible to run another test without navigating to another page
* Test that you can go back / forward in the browser history

### DIFF
--- a/src/app/components/domain/domain.component.ts
+++ b/src/app/components/domain/domain.component.ts
@@ -3,6 +3,7 @@ import {DnsCheckService} from '../../services/dns-check.service';
 import {ActivatedRoute} from '@angular/router';
 import {AlertService} from '../../services/alert.service';
 import { TranslateService } from '@ngx-translate/core';
+import { Location } from '@angular/common';
 
 @Component({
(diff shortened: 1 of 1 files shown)


pnax (approved):

mattias-p:
I tested this for v2021.2 and I like it a lot :)

matsduf:
@mattias-p, can you create an issue on this?

blacksponge:
@mattias-p I yes I am aware of that behaviour. This is because the view after a domain check and when navigating directly to the result are not the same (the domain check form is kept when we just tested a domain but not displayed if we go directly to a result). This is supposed to be a feature and is implemented as one.

mattias-p:
I created #284 to cover the back-and-forth misbehavior.

Apart for that this looks fine for v2021.2.

PR #258, opened 2021-08-04
Update URL to display the test hash ID

### LINKED ISSUE #240: Update the URL to display the test hash id
When a test completes, we currently stay on the same page and display the results. It would be nice to update the page URL with the unique URL associated to this test (the same as under the "Share" button) so that it is easy to copy/paste (and share) the results.
Doing this might also give the possibility to navigate through the browser history as the URL is updated between each test.

pnax:
## Purpose

This PR updates the URL once a test is finished. The new URL directly points to the test results.

## Context

Fixes #240 

## Changes

Update the URL once a test is finished. Use the [Location service](https://angular.io/api/common/Location).

## How to test this PR

Run a test and wait for it to finish. Once its finished and that you see the results, check the browser address bar, it should contain the hash ID.
It should also be possible to copy/paste this URL into another window and get the same results.

### DIFF
--- a/src/app/components/result/result.component.ts
+++ b/src/app/components/result/result.component.ts
@@ -1,3 +1,4 @@
+import { Location } from '@angular/common';
 import { Component, OnInit, OnChanges, Input, ElementRef, ViewChild } from '@angular/core';
 import {ActivatedRoute, Params} from '@angular/router';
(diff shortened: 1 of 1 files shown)


pnax:
Duplicate of #257

PR #252, opened 2021-07-21
Serve GUI from custom base url

blacksponge:
## Purpose

Make the GUI capable of being served from a custom base URL.

## Context

Addresses #250 

## Changes

* Fixes some dynamically built URLS.
* Add documentation for a custom installation

## How to test this PR

* Change the API endpoint to be `/zonemaster/api` in the prod environment file
* Build the app with `npm run build -- --base-href "/zonemaster/"`
* Serve the statics file using the following configuration (apache)
  ```apache
  <VirtualHost *:80>
    DocumentRoot /path/to/zonemaster-gui/dist
    DirectoryIndex index.html
    ServerAdmin example@example.net
    ServerName localhost
  
    <Directory "/path/to/zonemaster-gui/dist">
      Options Indexes MultiViews FollowSymlinks Includes
      AllowOverride All
      Require all granted
    </Directory>
  
    ProxyPass /zonemaster/api http://localhost:5000/
    ProxyPassReverse /zonemaster/api http://localhost:5000/
    ProxyPreserveHost On
    Alias "/zonemaster" "/path/to/zonemaster-gui/dist"
  
  </VirtualHost>
  ```
* Check that the GUI is loading and can access the API.
* Check that the FAQ pages are loading
* Check that the "share URL" is correct and working
* Repeat the steps with a normal build (default settings)

Issue #250, opened 2021-07-09
Run from sub directory


Ressy66:
For the life of me I can not get this thing to run from a sub directory.

ie:
example.net/zonemaster/
example .net has stuff in there that needs to be served, we have a speed test and other tool, we used to run the old ruby thing in example.net/zopnecheck after lots of stuffing around re aliaseing it worked for years, but nothing we do will see zponemaster run from it, in fact it wants to overtake our existing content, I wont go into WTF it is so overly unnecessarily complex to install ina website, thats as frustrating as hell for non expert web/js experts, so I implore devs to keep us simpletons in mind when you advance your code in years to come.

ok, so anyone have any ideas?  
I have tried changing rewritebase from /  to /zonemaster/, tried different combos of the conditions. no matter what or where I put it, it expects  all files are in  /   so all we get is strings of 404's when we try us   /vhosts/example.net/html/zonemaster/

Hope someone can help.
cheers

matsduf:
@Ressy66, if I understand it correctly, you want to run Zonemaster-GUI from e.g. https://example.com/zonemaster and 
https://example.com/zonemaster/domain_check instead of https://example.com/ and 
https://example.com/domain_check

You could easily create a subdomain for Zonemaster and run it from there, e.g. https://zonemaster.example.com/ and 
https://zonemaster.example.com/domain_check. Will that work for you?

Ressy66:
We are trying to avoid subdomain wastage and keep our "tools area" under one domain so its all in one spot.  
This is why we chose to run it from a sub directory. I just dont get why this was written to make it so impossible for average system admin to do anything with other than do what its authors want, if I wanted that I'd be running microsoft servers, not linux :)

I have managed to get it to  run, and using apache aliasing I can get /result/.....   now,  but no matter what aliasing I try the faq just does not show up -  thats not overly important, but still, its a something thats broken, and I dont like it when things are not normal.

Also, I would now be hesitant to ever upgrade the GUI for risk of having to start fixing it all over again.

matsduf:
We can take a look at it to see if it is possible to do what you request without too much changes. If you could provide us with what you have done to make it almost work it could help. If your installation is publicly available, please send the link. If you do not want to publish it here, you can go to my Github profile to find my email address.

matsduf:
> Also, I would now be hesitant to ever upgrade the GUI for risk of having to start fixing it all over again.

[...]

Issue #219, opened 2021-06-14
Configuration feature


mattias-p:
We should implement a configuration feature in GUI that allows the administrator to adapt how the GUI works in certain ways.

matsduf:
I guess that #170 could be resolved with such a configuration.

matsduf:
Resolved by #260

PR #260, opened 2021-08-04
add runtime configuration support

### LINKED ISSUE #170: Make GUI language selection dynamic
Backend has a method to report supported languages, "[get_language_tags]". There are three possible situations:
* Backend reports the same languages as GUI supports.
* Backend does not include all languages that GUI support, and then the excluded languages should not be included by GUI.
* Backend includes languages not supported by GUI.
  * One possibility is that those are ignored.
  * Another is that those are included, but GUI is still presented in English.

Today GUI has English as the default language, but Backend supports excluding English. That situation must be thought of.

----

See #219 for another way to solve this issue.

[get_language_tags]: https://github.com/zonemaster/zonemaster-backend/blob/master/docs/API.md#api-method-get_language_tags

blacksponge:
## Purpose

Add a runtime configuration file editable after compilation. This is useful to change the app settings (such as the api endpoint) without recompiling the whole thing.

## Context

* Suggested in #219
* Can fix #170 (current configuration make it easy to change the default language / restrict only to certain languages)
* Can make #252 / #250 easier

## Changes

* Add a configuration file `assets/app.config.json` that can be retrieve by the client. This configuration overrides the values of `environment.ts`, so we can treat the `environment.ts` values as default.
* Make the languages a configuration item instead of being hardcoded.

## How to test this PR

* As-is the GUI should behave as usual.
* Change the content of `assets/app.config.json` to be
  ```json
  {
    "defaultLanguage": "fr",
    "enabledLanguages": [ "en", "fr" ],
    "apiEndpoint": "https://zonemaster.fr/api"
  }
  ```
* Restart the GUI, verify that the changes reflect on the GUI.
* Other configuration items can tested, such as the logo or contact address.


pnax:
Great feature! This makes adjustments so much easier.
Could you provide documentation on how to configure each items (what about in a new file "docs/Configuration.md")?

src/assets/app.config.json
```diff
@@ -0,0 +1 @@
+{ }
```
pnax, line 1: Do we need to have both files "app.config.example.json" and "app.config.json"? Why not only keeping the last one?

blacksponge: When I will have written the documentation we can remove it, it is mostly to show case the feature.

matsduf:
>     * Change the content of `assets/app.config.json` to be
>       ```json
>       {
>         "defaultLanguage": "fr",
>         "enabledLanguages": [ "en", "fr" ],
>         "apiEndpoint": "https://zonemaster.fr/api"
>       }
>       ```

Is the `apiEndpoint` correct? In the documentation the example is "/api". Should it really be a full URL?

docs/Configuration.md
```diff
@@ -0,0 +1,22 @@
+# Configuration
+
+The GUI can be configured by changing the editing the file `assets/app.config.json`.
+
+The supported configuration items are the following.
+
+* `"apiEndpoint"`: The URL to use to contact the API, default `"/api"`.
+* `"defaultLanguage"`: The default language of the GUI, defined as a two letters
+  code, default `"en"`. Accepted values: `"da"`, `"en"`, `"fi"`, `"fr"`,
+  "`nb"`, `"sv"`.
```
matsduf, line 10: Can the `defaultLanguage` really be a language outside the `enabledLanguages`?

docs/Configuration.md
```diff
@@ -0,0 +1,22 @@
...
+* `"enabledLanguages"`: An array of the languages enabled in the GUI, default
+  `[ "da", "en", "fi", "fr", "nb", "sv" ]`.
+* `"contactAddress"`: The contact email address displayed in the footer, default
```
matsduf, line 13: Instead of listing the languages we should refer to the new common.ts. I guess that is true for all values below too.

matsduf (changes requested):
`docs/TranslationGuide.md` should also be updated.

docs/Configuration.md
```diff
@@ -0,0 +1,22 @@
...
+  `"contact@zonemaster.net"`.
+* `"logo`": The URL to the image displayed in the navigation bar, default
+  `"assets/images/zonemaster_logo_2021_color.png"`.
```
matsduf, line 16: The Zonemaster logo is part of the GUI. I see no reason why it should be possible to change that with a simple configuration. If you want to change that, do a recompilation.

docs/Configuration.md
```diff
@@ -0,0 +1,22 @@
...
+* `"clientInfo"`: An object containing the client information reported to the
+    API. The object must have the following properties:
+    * `"version"`: The GUI version, default to the current Zonemaster GUI
+      version. The value is also displayed along the other components version
+      on the *program versions* tooltip.
+    * `"id"`: The name of the client, default to `"Zonemaster GUI"`.
```
matsduf, line 22: The configuration is of a compiled GUI. The version does not change by reconfiguration. And it is still "Zonemaster GUI" (which should probably be "Zonemaster-GUI"). There is no reason to include "clientInfo" in the configuration file.

docs/Configuration.md
blacksponge: No only the languages defined in `enabledLanguages`, I should word that differently.

docs/Configuration.md
blacksponge: I'd rather not point the user to a source file and instead have the documentation being self contained.

docs/Configuration.md
blacksponge: I am not sure, it is common for other software to allow for easy customization of the GUI. I don't see the point for *forcing* a recompilation when the option is already available, documented or not.

docs/Configuration.md
blacksponge: I could remove that from the doc.

docs/Configuration.md
matsduf: I see your point and that means that we have to update this file too when adding or deleting a language. It should then be listed in `docs/TranslationGuide.md`.

docs/Configuration.md
matsduf: You mean that you could just replace the logo file. That is correct, so OK, keep it.

docs/Configuration.md
matsduf: It should be removed from the configuration file too, as I see it.

src/environments/common.ts
```diff
@@ -0,0 +1,18 @@
+// Common values for all environments
+// You can override those in environment configuration files
+
+export const common = {
+  contactAddress: 'contact@zonemaster.net',
+  logoUrl: 'assets/images/zonemaster_logo_2021_color.png',
+  clientInfo: {version: '3.3.0', id: 'Zonemaster GUI'},
```
matsduf, line 7: The `clientInfo` should not be possible to override. That is the version and identity of the code. The rest is fine.

docs/Configuration.md
```diff
@@ -0,0 +1,15 @@
+# Configuration
+
+The GUI can be configured by changing the editing the file `assets/app.config.json`.
+
+The supported configuration items are the following.
+
+* `"apiEndpoint"`: The URL to use to contact the API, default `"/api"`.
```
matsduf, line 7: Could this either be a full URL, such as "https://zonemaster.fr/api", or the path part, such as "/api"? Please clearify.

docs/Configuration.md
blacksponge: I have updated the translation guide to include the new instructions

docs/Configuration.md
blacksponge: I moved it to a dedicated module

docs/Configuration.md
blacksponge: done

src/environments/common.ts
blacksponge: done

docs/Configuration.md
blacksponge: It could be either. Without the origin it would use the same origin as the one serving the GUI (as it is with the example apache configuration). We were already using both, https://zonemaster.net/api for the dev / testing environment and `/api` for prod. I tried to clarify a bit the documentation to make that clearer.

pnax (dismissed):

docs/TranslationGuide.md
```diff
@@ -127,16 +131,31 @@ import 'moment/locale/xx';
...
 ```
-and add the two-letter language code of the new language. Preserve
-the alphabetical order of the language codes.
+and append the new two-letter language code of the new language.
+
+
+
+### Configuration.md
+
+Add the new languge's two-letter code to the list of default values for
```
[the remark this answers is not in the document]
blacksponge: fixed

docs/Configuration.md
```diff
@@ -0,0 +1,19 @@
+# Configuration
+
+The GUI can be configured by changing the editing the file `assets/app.config.json`.
```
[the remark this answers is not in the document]
blacksponge: fixed

matsduf (approved):

mattias-p:

PR #261, opened 2021-08-19
wait for the lang nav before clicking

blacksponge:
## Purpose

Make running e2e tests more reliable.

Now that the language nav is populated dynamically from the configuration, the tests can fail as the elements to change the language do not exists at the time of the click.

## Changes

Try to wait for the languages list to be loaded.

## How to test this PR

Run the e2e tests

### DIFF
--- a/e2e/utils/app.utils.ts
+++ b/e2e/utils/app.utils.ts
@@ -10,7 +10,8 @@ export class Utils {
   }
 
   setLang(lang) {
-    return element(by.xpath('//a[@lang="' + lang + '"]')).click()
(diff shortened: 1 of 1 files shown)


blacksponge:
Travis builds are not starting because ` Owner zonemaster does not have enough credits.`

matsduf (approved):
Does this PR resolve any issue or problem? It is not clear from the description.

blacksponge:
> Does this PR resolve any issue or problem? It is not clear from the description.

I updated the description. I have seen the problem on other pull requests where I had to force restart the build a couple times for the tests to pass.

blacksponge:
An example of that issue can be seen on this build: https://app.travis-ci.com/github/zonemaster/zonemaster-gui/builds/237168891

Issue #115, opened 2020-04-21
Line break on non-wide windows


matsduf:
When a test is presented the following three elements are presented on the same line:

* Test number, e.g. 'Test # 8'
* Domain name tested, e.g. 'nic.vermögensberatung'
* Four boxes, buttons:
  - "History"
  - "Export"
  - "Share"
  - "Settings"

If the domain namn is long or if the language is not English, especially if it is French or Danish, the line is broken in an ugly way. Also depending on window width. I suggests that the four boxes are displayed on a line above, but still to the right, to avoid line breaks.

![image](https://user-images.githubusercontent.com/5430733/79852749-5cfc7180-83c7-11ea-84ac-abc5e5bd37fe.png)

![image](https://user-images.githubusercontent.com/5430733/79852815-769db900-83c7-11ea-8622-3a71e62e8815.png)

![image](https://user-images.githubusercontent.com/5430733/79852873-90d79700-83c7-11ea-8a47-89f4c99a38a8.png)

![image](https://user-images.githubusercontent.com/5430733/79852933-a6e55780-83c7-11ea-9f07-a86b443b25ed.png)

pnax:
Resolved with #263

PR #263, opened 2021-08-23
Improve result UI

blacksponge:
## Purpose

Improve overall result UI

## Context

* Addresses an horizontal overflow issue on mobile screens
* Fix navigation bar animation
* Addresses #115

## Changes

* Dependencies upgrade (angular, typescript, ngx-translate, ng-bootstrap)
* Make the module header sticky
* Refactor result component HTML (tables do not support sticky element very well)
* Hide the module column on mobile device (actually I wonder if this column is still useful on wide screen as well now that the module header is sticky, any thoughts ?)
* Fix the dimension of the github thingy
* Fix the navigation bar animation
* Fix "back to top" button
* Set a fixed height (80px) for the nav bar on wide screens
* Make the generation of the HTML export (almost) independent from the component rendering (so that changing the component template don't affect the export).
* Attempt to fix blurry logo on Chrome
* Fix the filters / action buttons alignment on narrow devices (#115)
* Remove the "ungroup" feature
* Remove the header of the result table

## How to test this PR

* Make a test
* Expend some modules in the result, enough to scroll
* Check that the module name is sticky, just below the nav bar on both mobile and desktop screens
* Clicking the sticky header should collapse the module but the header should not move because of the scroll offset when collapsing the module
* The module column should not be visible on narrow devices
* There should be no horizontal overflow on any devices


pnax:
Nice work.

> Hide the module column on mobile device (actually I wonder if this column is still useful on wide screen as well now that the module header is sticky, any thoughts ?)

When disabling the setting "Group by module", the module name does not appear anymore. Maybe we should decide if we want to keep this setting to ungroup the results.
And to answer your question, I don't really know I would say that by removing this column on wide screen the header might be far from the line (especially in big groups such as DNSSEC). So I would argue against removing it. But to be honest this is just a feeling based on my habits.

src/app/components/navigation/navigation.component.css
```diff
@@ -13,42 +19,50 @@
...
-	 .github {
-		         display: none
-			  }
-		  }
+  .github {
+    display: none
+  }
+
+  nav {
+    height: auto;
```
pnax, line 58: It seems that a fixed height is used everywhere else. Why using an `auto` height here? TBH I am working on a information banner and I need the nav height to be always fixed with a known value. Would it be possible to let the nav height be 80px here as well?

blacksponge: >  Would it be possible to let the nav height be 80px here as well?

It is not possible to fix the nav height on devices smaller than 992px because the navigation bar switch to a drop down menu. The height is then variable in that case, hence making it auto to not break the menu.

Screenshot 2021-08-25 at 16-12-31 Zonemaster

If you need to know the height of the navigation bar, this PR adds a new service `NavigationService` because i had the same problem. This service exposes a observable that you can subscribe to to update a fixed height somewhere.

Relevant files:
https://github.com/zonemaster/zonemaster-gui/blob/e246b4571079577c311e1a6efeb15da1d89da189/src/app/services/navigation.service.ts#L4-L10

https://github.com/zonemaster/zonemaster-gui/blob/e246b4571079577c311e1a6efeb15da1d89da189/src/app/components/result/result.component.ts#L89-L91

https://github.com/zonemaster/zonemaster-gui/blob/e246b4571079577c311e1a6efeb15da1d89da189/src/app/components/result/result.component.html#L159

blacksponge:
> When disabling the setting "Group by module", the module name does not appear anymore. Maybe we should decide if we want to keep this setting to ungroup the results.

Oh I didn't though of that. I don't see the use case of not grouping by module so I am okay to remove it. It was used before for the generation of the html export but this PR remove that coupling between the view and the export making it safe to remove it.

blacksponge:
>  Maybe we should decide if we want to keep this setting to ungroup the results.

In the end I removed the feature.

matsduf:
> > Maybe we should decide if we want to keep this setting to ungroup the results.
> 
> In the end I removed the feature.

I see no point to keep it.

matsduf (dismissed):
I have tested in a computer by making the window narrow, and it looks fine.

matsduf:
@blacksponge, can you fix the conflict?

matsduf (approved):

matsduf (approved):
This is fine.

pnax (approved):
Nice improvements!

matsduf:
@blacksponge, please merge.

pnax:
## Testing v2021.2


- [x] Check that the module name is sticky, just below the nav bar on both mobile and desktop screens
- [x] Clicking the sticky header should collapse the module but the header should not move because of the scroll offset when collapsing the module
- [x] The module column should not be visible on narrow devices
- [x] There should be no horizontal overflow on any devices

LGTM.

PR #264, opened 2021-08-24
Fix favicon images

### LINKED ISSUE #119: Distorted favicon
The favicon used for the GUI is distorted. It should have the same proportions as that part of the logotype.

blacksponge:
## Purpose

Fix the favicons size

## Context

Fix #119

## Changes

Regenerate all favicons so that the logo is not distorted.

## How to test this PR
The logo should have the correct ratio size in all those cases : (I can not test on max / ios)

* Load the GUI, check the favicon
* Add the website to the visited sites / shortcuts in firefox
* Add the website to the visited sites / shortcuts in chrome
* Add the website to the windows 10 start new
* Add the website to the android home screen


matsduf (approved):
The favicon looks fine on Firefox, Safari and Chrome on MacOS. I have both in tab and in the bookmark list. There is one exception, on Firefox the favicon is still distorted (horizontally compressed) in the bookmark list, but I think we can live with that for now.

blacksponge:
> on Firefox the favicon is still distorted (horizontally compressed) in the bookmark list

Could it be cached?

matsduf:
> Could it be cached?

That could be the case. I do not see it as a show stopper, so that is why I approved the PR.

pnax:
## Testing v2021.2


- [x] Load the GUI, check the favicon
- [x] Add the website to the visited sites / shortcuts in firefox
- [x] Add the website to the visited sites / shortcuts in chrome
- [x] Add the website to the windows 10 start new
- [ ] Add the website to the android home screen - cannot test this OS

matsduf:
I suggest that consider this to enough tested as of now, and after release we can see the behavior on Chrome on Android, which is not a supported platform.

PR #265, opened 2021-08-24
unsubscribe on component ondestroy

### LINKED ISSUE #262: Language change event not cleared when changing route
When a component that watches the language changes (like the result or faq components) is loaded and then "unloaded" by navigating to a different route (like /domain_check) the event is never cleared. As a result changing the language again after visiting a result or the faq will still result in queries being made even though the event should not be watched anymore.

Step to reproduce :
1. Visit https://zonemaster.net/result/40047b0ede87b43a
2. Go to the faq page
3. Go to the domain check page
4. Change the language
5. By opening the dev tools you can see query to the api (`get_test_results`) and `gui-faq-{lang}.html` being made.

blacksponge:
## Purpose

Fix subscription leaks

## Context

Fixes #262

## Changes

Unsubscribe needed subscription when the component is destroyed.

## How to test this PR

* Changing the language after visiting the FAQ while not being on the FAQ page should not result in the application querying the FAQ.
* Same thing for a test result.

### DIFF
--- a/src/app/components/faq/faq.component.ts
+++ b/src/app/components/faq/faq.component.ts
@@ -1,22 +1,26 @@
-import { Component, OnInit, AfterViewChecked } from '@angular/core';
+import { Component, OnInit, AfterViewChecked, OnDestroy } from '@angular/core';
 import {TranslateService, LangChangeEvent} from '@ngx-translate/core';
 import {ActivatedRoute} from '@angular/router';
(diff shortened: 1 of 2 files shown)


matsduf (approved):

mattias-p:

While testing I found #282, but I think that's out of scope for this PR.

PR #266, opened 2021-08-25
Information banner

pnax:
## Purpose

This adds a banner that can be used to display a custom message to the users.
FYI I'm no expert with colors, I choose one I felt was fine to my eyes.

zm-gui-info-1

zm-gui-info-2

## Context

N/A

## Changes

* Update the CSS to make the main component just below the navbar

* Add a new configuration item `msgBanner` that can be used to display a custom message on top of the page below the navbar

## How to test this PR

Set the `msgBanner` configuration item with some text. Reload the page, a banner with the text should be displayed. Removing the item from the configuration or setting an empty string should remove the banner.


blacksponge:
> Such a message is outside Zonemaster as such. That is something for the system that hosts Zonemaster. We are already discussing an update of the GUI with the help of outside resources. I suggest that we do not add this feature.

I think this is a useful feature, and I don't see why it should not be part of the GUI. We needed it for communicating about upcoming maintenance on zonemaster.net, and it can be used to communicate on other things (like a survey if we do one). I understand that we are working towards more GUI customization, and that this feature could become obsolete in the future. But in the mean time, I think we should add this as it is already an improvement and we would make use of it.

pnax:
> Maybe the color of the banner can be improved though, for reference if you want to use bootstrap's alert colors: https://getbootstrap.com/docs/4.0/components/alerts.

Thanks for the pointer. I used it.

I also added support of HTML within the banner message.

> If there is an upcoming service it could be valuable for a very frequent user, but if the users are infrequent there is not much use for the information. If it is important to inform about service stop then there should be a mechanism to give information during service stop, and that is outside Zonemaster.

Being able to communicate with people using the service is not something out of interest I think. This banner does not need to be present all the time. This is just to provide an easy way to reach out to the people using the service.

I do think this adds value to the GUI. And since this PR is ready, I don't see the point in waiting after next release to merge it (I am completely biased since this is a PR I made).

matsduf (changes requested):
It works as specified, but `dist/assets/app.config.sample.json` should also be updated.

pnax:
> but dist/assets/app.config.sample.json should also be updated

I'm not sure if it _should_ since I made this property optional but I can add the entry to the file anyway.

matsduf (dismissed):

pnax:
rebased on top of develop to make unhappy Travis happy. Please re-review.

matsduf (approved):
I approve.

mattias-p:

PR #267, opened 2021-08-26
don't overwrite the config file on update

blacksponge:
## Purpose

When copying the new files after an update the configuration could be overwritten as is it present in the source.

## Changes

This PR moves the configuration out of the source and provides instruction to create the file.

## How to test this PR

* Build the GUI and make an archive from the dist directory
* Unarchive the archive to a temporary directory
* Create and edit the configuration
* Unarchive the archive to the same directory
* The configuration should remain untouched


docs/Configuration.md
```diff
@@ -1,6 +1,12 @@
 # Configuration
 
-The GUI can be configured by editing the file `assets/app.config.json`.
+The GUI can be configured by creating the file `assets/app.config.json`.
```
matsduf, line 3:
```
The GUI can be configured by creating the file `DISTBASE/dist/assets/app.config.json`. "DISTBASE"
is where the Zonemaster Web GUI zip file is installed, which is `/var/www/html/zonemaster-web-gui`
by default installation (see [Installation instructions](Installation.md)).
```

docs/Configuration.md
```diff
@@ -1,6 +1,12 @@
...
+
+This file can by created by copying `assets/app.config.sample.json` to `assets/app.config.json`:
```
matsduf, line 5:
```
by copying `app.config.sample.json` found in `DISTBASE/dist/assets`:
```

docs/Configuration.md
```diff
@@ -1,6 +1,12 @@
...
+```sh
+cp assets/app.config.sample.json assets/app.config.json
```
matsduf, line 7:
```
cp app.config.sample.json app.config.json
```

matsduf (changes requested):
To be helpful, src/assets/app.config.sample.json could be populated with the default values to be a real example instead of just an empty file.

matsduf:
Another alternative would be to suggest, in the installation instructions, that any `app.config.json`is saved before upgrading GUI. In that way we could suggest that the "DISTBASE" is removed to ensure that old files do not interfere with a new installation.

blacksponge:
> To be helpful, src/assets/app.config.sample.json could be populated with the default values to be a real example instead of just an empty file.

I put the default values in the file and updated the documentation.

>  Another alternative would be to suggest, in the installation instructions, that any app.config.json is saved before upgrading GUI. In that way we could suggest that the "DISTBASE" is removed to ensure that old files do not interfere with a new installation.

I prefer to not have the sample / default configuration tracked in git anyway, so that we can edit it while keeping a clean tree.

matsduf:
> I prefer to not have the sample / default configuration tracked in git anyway, so that we can edit it while keeping a clean tree.

Yes, a sample config file should just be a sample. I think the change is good. I think we should add a comment in the installation instruction to save any `app.config.json` before upgrading, but that does not have to be included in this PR.

docs/Configuration.md
```diff
@@ -1,6 +1,17 @@
 # Configuration
 
-The GUI can be configured by editing the file `assets/app.config.json`.
+The GUI can be configured by creating the file `DISTBASE/dist/assets/
+app.config.json`. "DISTBASE" is where the Zonemaster Web GUI zip file is
```
matsduf, line 4: `DISTBASE/dist/assets/app.config.json` should be on the same line or else it is rendered with a space:
```
DISTBASE/dist/assets/ app.config.json
```

matsduf (changes requested):

docs/Configuration.md
blacksponge: fixed

matsduf (dismissed):

blacksponge:
Please re-review, I have made a small update to ignore the config file when running the tests.

src/environments/environment.tests.ts
```diff
@@ -4,5 +4,6 @@ export const environment = {
   ...common,
   production: false,
   apiEndpoint: 'https://zonemaster.net/api',
-  mock: true
+  mock: true,
+  configUrl: 'assets/app.config.tests.json'
```
matsduf, line 8: I cannot see that the file 'assets/app.config.tests.json' exists. Should it be 'assets/app.config.sample.json'? Or is the other file created some way?

matsduf (changes requested):

src/environments/environment.tests.ts
blacksponge: I use a non existing file on purpose, so that the config used is always the default (`environment.tests.ts`). I can document that in this file if needed.

matsduf: Yes, please do. Or maybe rename the file to 'assets/app.config.non-existent.json' to make it more obvious that it should not exist.

blacksponge: Done

matsduf (approved):

matsduf:
@blacksponge, can you push an empty commit to see if Travis will test it properly?

matsduf:
@blacksponge, has anything been changed by your force-push?

blacksponge:
> has anything been changed by your force-push?

No I just rebased on top of develop

matsduf (approved):

matsduf:
@blacksponge, has anything been changed by your force-push?

blacksponge:
Rebased again to fix the tests

matsduf (approved):

mattias-p:

Issue #239, opened 2021-07-05
Dynamic server validation error translation


blacksponge:
#230 and https://github.com/zonemaster/zonemaster-backend/pull/853 add server validation error. As stated in the discussion following https://github.com/zonemaster/zonemaster-gui/pull/230#issuecomment-874123079 the message are not translated if the language is changed after the validation occurs.

This should be addressed.

blacksponge:
resolved by #269

PR #269, opened 2021-10-14
Translate backend validation errors

blacksponge:
## Purpose

Translate backend validation errors

## Context

Depends on https://github.com/zonemaster/zonemaster-backend/pull/891, addresses #239

## Changes

* Add language in param for start_domain_test and fetch_data_from_parent_zone
* Resend form if the form has been touched by the user and is in a invalid state

## How to test this PR

* Fill garbage in the test form and submit it.
* Check that errors are displayed in the selected language (if supported by the backend)
* Change the language
* Check that the errors have also been translated

### DIFF
--- a/src/app/components/form/form.component.ts
+++ b/src/app/components/form/form.component.ts
@@ -60,7 +60,7 @@ export class FormComponent implements OnInit, OnChanges, OnDestroy {
   ngOnInit() {
     this.langChangeSubscription = this.translateService.onLangChange.subscribe((_: LangChangeEvent) => {
       if (this.newForm.touched) {
-        this.runDomainCheck(true);
+        this.runDomainCheck(false);
       }
     });
     this.generate_form();
@@ -233,7 +233,7 @@ export class FormComponent implements OnInit, OnChanges, OnDestroy {
     }
   }
 
-  public runDomainCheck(submitNotValid = false) {
+  public runDomainCheck(submitValid = true) {
     this.newForm.markAllAsTouched();
     let param = this.newForm.value;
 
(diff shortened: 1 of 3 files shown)


matsduf (approved):
In general this looks good.

There are some cases where you change language when you get a language mix, but I think we can live with it as of now.

I approve.

matsduf (approved):

src/app/components/form/form.component.ts
```diff
@@ -221,7 +233,7 @@ export class FormComponent implements OnInit, OnChanges {
     }
   }
 
-  public runDomainCheck() {
+  public runDomainCheck(submitNotValid = false) {
```
mattias-p, line 236: Maybe give the argument a positive name?

src/app/components/form/form.component.ts
```diff
@@ -250,7 +262,7 @@ export class FormComponent implements OnInit, OnChanges {
         digest: ds.digest
       }});
 
-    if (this.newForm.valid) {
+    if ((!submitNotValid && this.newForm.valid) || (submitNotValid && !this.newForm.valid)) {
```
mattias-p, line 265: Could this be expressed as `submitNotValid != this.newForm.valid`?

mattias-p (dismissed):

matsduf (dismissed):
I approve this, and will re-approve if the comments by @mattias-p are considered.

src/app/components/form/form.component.ts
blacksponge: I inverted the state so it has a positive name

src/app/components/form/form.component.ts
blacksponge: changed

matsduf (approved):
I approve.

mattias-p (approved):

mattias-p:
I tested this for v2021.2. It works for Swedish and French, which are the only ones currently supported in Backend. So, LGTM.

Issue #277, opened 2021-11-25
Some placeholders are not properly translated


pnax:
While testing #227, I found that the "Algorithm" and "Digest type" placeholders are not properly translated.

1. open the options
2. change the language
3. see all placeholders be translated except for the "Algorithm" and "Digest type"

Actually both are not placeholders:

https://github.com/zonemaster/zonemaster-gui/blob/867f12cd796e519d6531137af37c9317c6f6d976/src/app/components/form/form.component.html#L160

https://github.com/zonemaster/zonemaster-gui/blob/867f12cd796e519d6531137af37c9317c6f6d976/src/app/components/form/form.component.html#L185

matsduf:
Can you see if the problem is browser dependent?

pnax:
Indeed, it seems to be browser dependent. A fix would be to remove the "hidden" property.

matsduf:
In what browser do you get the problem, and not get the problem, respectively? What effect would removing have?

pnax:
I tested it with Firefox. And in the code, I can see the translation. It is just not rendered.
See #278

pnax:
Closed by #278

PR #278, opened 2021-11-25
Properly translate all placeholders

pnax:
## Purpose

Some placeholders are not properly translated (due to limitation from the browser). Removing the hidden HTML property would fix this issue.

## Context

Addresses #277 

## Changes

Remove the hidden property.

## How to test this PR

Repeat this on Firefox, Chrome and mobile device:

1. open the options
2. change the language
3. check that the "Algorithm" and "Digest type" are properly translated


matsduf:
> ## How to test this PR

We should add, repeat this on Firefox and Chrome. Optionally also on mobile devices.

blacksponge (approved):

matsduf (approved):

Issue #118, opened 2020-04-21
Change default on displayed message levels


matsduf:
Today, the GUI will display messages on levels by default, from INFO to CRITICAL. In most cases the high number of messages on INFO level will hide other messages, and for most users INFO messages are probably not in focus. I suggest that the default is changed so that INFO messages are only shown when they have been selected.

The CLI shows NOTICE and higher by default.

Updated 2022-01-25:

Since the messages are not shown by default the in the current design, instead the levels (Basic, DNSSEC etc) are shown, and those display the severity level by color of the message in the level with the highest severity level, it is an acceptable work-around that the messages are sorted by severity level with the highest level highest per level (Basic, DNSSEC etc).

pamasse:
You suggest showing only NOTICE/WARNING/ERROR/CRITICAL messages and not ALL ?
Basically, it's like selecting only the NOTICE/WARNING/ERROR/CRITICAL filter button?

matsduf:
Yes.
Yes. I suggest that to be the default, but not now. This is a proposal for v2020.1.

matsduf:
Resolved by #297

PR #297, opened 2022-01-24
Sort messages by severity

blacksponge:
## Purpose

Show higher severity level messages first.

## Context

Suggested as a workaround for issue #118

## Changes

The message or sorted in place by severity before being displayed.

## How to test this PR

Test a domain that output different message with different severity level.

Output bellow generated by testing `ppsfleet.navy` in undelegated mode, the data was fetched from the nameserver, `ns3.ppsfleet.navy` removed from the ns list and `fake.ppsfleet.navy` pointing to 127.0.0.1 was added.

Screenshot 2022-01-24 at 16-12-23 Zonemaster

### DIFF
--- a/src/app/components/result/result.component.ts
+++ b/src/app/components/result/result.component.ts
@@ -164,7 +164,7 @@ export class ResultComponent implements OnInit, OnChanges, OnDestroy {
 
       for (const module in this.module_items) {
         this.module_items[module].sort((msg1, msg2) => {
-          // sort messages by descending serverity level
+          // sort messages by descending severity level
           return this.levelSeverity.indexOf(msg2.level) - this.levelSeverity.indexOf(msg1.level);
         })
       }
--- a/src/app/components/result/result.component.ts
+++ b/src/app/components/result/result.component.ts
@@ -162,6 +162,13 @@ export class ResultComponent implements OnInit, OnChanges, OnDestroy {
         this.level_items[item['level'].toLowerCase()].push(item);
       }
 
+      for (const module in this.module_items) {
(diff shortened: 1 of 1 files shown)


mattias-p (approved):
I haven't tested but the code looks right. I guess the time complexity isn't great but the data sets are small so that shouldn't matter.

matsduf:
Is the sorting done in such a way that order within the same severity level is retained compare to unchanged GUI? As it is now, the messages come in the order of creation and that means that related messages come after each other.

blacksponge:
> Is the sorting done in such a way that order within the same severity level is retained compare to unchanged GUI? As it is now, the messages come in the order of creation and that means that related messages come after each other.

According to the `sort` documentation, it should keep the original order. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#sort_stability

matsduf (dismissed):

src/app/components/result/result.component.ts
```diff
@@ -162,6 +162,13 @@ export class ResultComponent implements OnInit, OnChanges, OnDestroy {
         this.level_items[item['level'].toLowerCase()].push(item);
       }
 
+      for (const module in this.module_items) {
+        this.module_items[module].sort((msg1, msg2) => {
+          // sort messages by descending serverity level
```
tgreenx, line 167: Serverity -> Severity

tgreenx (approved):

src/app/components/result/result.component.ts
blacksponge: fixed

matsduf (approved):
I approve.

Issue #275, opened 2021-11-25
Deprecated dependencies


blacksponge:
* Protractor (used for e2e testing) is not really maintained any more. I suggest we replace it with alternative like https://playwright.dev for example.
* Moment.js [has announced](https://momentjs.com/docs/#/-project-status/recommendations/) its end of development, we should either consider moving to something else, maybe [day.js](https://day.js.org)?

blacksponge:
Fixed by #299 and #298

PR #298, opened 2022-01-25
Migrate e2e tests to playwright

blacksponge:
## Purpose

I chose to not use Protractor for testing anymore. The library has became less and less maintained up to the point that other complementary libraries (that we used) have stop development. This impact our ability to update our dependencies as well as the minimum required node version, putting us at risk.

This PR migrate all e2e tests to Playwright. The framework is being maintained by Microsoft ensuring (I hope) that it will remain maintained for some time. Also it removes the need for a third party library to do visual comparison and the pre-dependency on Chrome.

## Context

 #275

## Changes

Update all tests.

I am also hoping for the tests to be more readable and less hacky.

## How to test this PR

If the CI is happy, it works. You can also test it on your environment but some tests for visual comparison might fail due to differences in rendering and timezone.


matsduf (dismissed):
This looks good and travis is happy.

blacksponge:
rebased on top of develop to ensure that the tests are still passing

pnax (approved):

PR #299, opened 2022-01-26
remove dependency on moment

blacksponge:
## Purpose

Remove the dependency on moment.js as the library as stopped development (and that we don't use it that much anyway).

## Context

#275

## Changes

* Parse date using standard javascript instead of moment.

## How to test this PR

* Navigate though the GUI, if no error are shown in the console and that nothing looks out of place it should be good.


pnax:
Testing this, I encounter what looks like a bug:

1. run a test
2. click on the "History" box (you should get the current test and previous one correctly listed)
3. close the box
4. open the "History" again <- no tests are listed anymore and the console is filled with the following error message:
    ```
    ERROR Error: InvalidPipeArgument: 'Unable to convert "Invalid Date" into a date' for pipe 'DatePipe'
    ```

blacksponge:
> Testing this, I encounter what looks like a bug:
> ...

This should be fixed now

pnax (approved):

PR #300, opened 2022-01-27
make navigation more consistent

### LINKED ISSUE #284: Browser history misbehavior
I found this funny behavior while release testing #257. I'm using Firefox 94 on Fedora 35.

When I go back and forth between the check domain page and the result page using my browser's back and forward buttons I get some weird behavior. It starts out a bit funny before it settles on a stable and reasonable pattern from step 7.

1. I open a new tab and go to http://192.168.56.7/
2. The URL bar is set to http://192.168.56.7/domain_check, and the page view shows a "check domain" component above an "about zonemaster" component.
3. I start a test for zonemaster.net
4. When the test is completed the URL bar is set to http://192.168.56.7/result/c020fda0379befc3, and the page view shows a "result component" below a "check domain" component.
5. I press my browser's Back button.
6. The URL bar is set to http://192.168.56.7/domain_check, and the page view stays unchanged.
7. I press my browser's Forward button.
8. The URL is set to http://192.168.56.7/result/c020fda0379befc3 again, and the page view shows a "result component" below a "Result heading" component.
9. I press my browser's Back button.
10. The URL bar is set to http://192.168.56.7/domain_check, and the page view shows a "check domain" component above an "about zonemaster" component.
11. I go back to step 7.

blacksponge:
## Purpose

Fix inconsistency issues on navigation.

## Context

Fixes #284

## Changes

* The domain checking part of the `domain` component is moved to a new `domain-check` component.
* The `domain` component is only used as a landing page and only display the `domain` component and the about section. No logic is implemented there.
* The `domain-check` component *actually* redirect to the `result` page (as opposed to only changing the url like it was doing before)
* The `result` component embed the `domain-check` component when it is accessed after a test finished
* The link in the navigation menu is no longer "active" when viewing result after being redirected from the form page

## How to test this PR

* Create a new test
* Verify that when the test finishes the browser is redirect to the result page with an url `/result/<id1>`. The domain check form is visible.
* Press the back button in the browser.
* Verify that only the domain check for is displayed. The url should be `/domain_check`.
* Press the forward button in the browser.
* Verify that the previous test result is displayed with the url `/result/<id1>`. The domain check form should still be visible.
* Create a second test for an other domain from the result page.
* When the second test finishes the url should change to `/result/<id2>`. The result for the second test should be displayed and the domain check form should still be visible.
* Refresh the page.
* The domain check form is not visible, instead a `Result` header is displayed.


pnax (approved):
nice!

PR #303, opened 2022-02-14
remove unused scripts and dependencies

blacksponge:
## Purpose

We have multiple tests environment but essentially only use one. This PR remove the dependencies and scripts that we don't use.

## Context

Follow up on discussion here https://github.com/zonemaster/zonemaster-gui/pull/302#issuecomment-1039039358

## Changes

* Remove karma and the hand-made e2e test script.
  For now I suggest we focus on the tests that we have, which should be improved as there as some stuff not covered, before we see what can be done in addition to those. Then if using karma make sense we could re-add it.
* Remove the local_test.js script
   It looks like most of what this test tested has been replaced by the e2e tests.

## How to test this PR

...


matsduf (approved):
This is a good approach.

blacksponge:
> do we still need this dependency, since it looks like the code related to it is removed?

Yes, good catch, I removed it

pnax (dismissed):

mattias-p:
I don't have a good understanding of what these things are, but one thing I noticed is that `local_test.js` seems to contain a test case. I'm not sure what it does. But I have two questions about it:
1. Is it a good test?
2. Is it covered by the remaining test suite?

If it is a good test that is not covered by the remaining test suite, maybe we could migrate it instead of simply removing it?

blacksponge:
> Is it a good test?
> Is it covered by the remaining test suite?

This is testing several things, several of them are already tested, some are outdated, some are overspecific in my opinion. The test case that may be relevant and that we are not testing already is testing the that the "fetch data from parent zone" button populates the fields correctly. I will create an issue of this one.

mattias-p:
> > Is it a good test?
> > Is it covered by the remaining test suite?
> 
> This is testing several things, several of them are already tested, some are outdated, some are overspecific in my opinion. The test case that may be relevant and that we are not testing already is testing the that the "fetch data from parent zone" button populates the fields correctly. I will create an issue of this one.

Sounds great :)

blacksponge:
rebased on top of develop

matsduf (approved):
Looks good.

Issue #249, opened 2021-07-12
Remove leading and trailing space


matsduf:
It could be discussed if and how much GUI should validate input data, but on point it is clear that it should be done and that is leading and trailing on data from input fields. Furthermore, such error should not be reported. Leading and trailing space should just silently be removed before anything is sent to backend. If GUI does any validation of fields, e.g. validates that zone to be tested is not empty, then that validation should be done after removal of leading and trailing spaces.

Currently GUI sends leading and traling spaces to Backend.

matsduf:
Fixed by #305

PR #305, opened 2022-02-15
Fix trailing dots spaces

blacksponge:
## Purpose

Remove trailing spaces and dots on domains

## Context

#249 

## Changes
* rename form variable (just for clarity)
* remove trailing spaces / dots and leading spaces

## How to test this PR
* submit a test with trailing spaces or dots in domain and / or NS
* check that the test does not crash
* the form inputs should contain the updated value

### DIFF
--- a/src/app/components/form/form.component.ts
+++ b/src/app/components/form/form.component.ts
@@ -235,7 +235,12 @@ export class FormComponent implements OnInit, OnChanges, OnDestroy {
 
   // Remove trailing spaces and dots, and leading spaces
   private sanitizeDomain(domain: string): string {
-    return domain.replace(/[ \.]+$/, '').trim();
+    domain = domain.trim();
+    if (domain == '.') {
+      return domain;
+    } else {
+      return domain.replace(/[ \.]+$/, '');
+    }
   }
 
   public runDomainCheck(submitValid = true) {
--- a/src/app/components/form/form.component.ts
+++ b/src/app/components/form/form.component.ts
@@ -233,9 +233,17 @@ export class FormComponent implements OnInit, OnChanges, OnDestroy {
     }
   }
 
+  // Remove trailing spaces and dots, and leading spaces
+  private sanitizeDomain(domain: string): string {
+    return domain.replace(/[ \.]+$/, '').trim();
+  }
+
   public runDomainCheck(submitValid = true) {
     this.form.markAllAsTouched();
     let param = this.form.value;
+
+    param.domain = this.sanitizeDomain(param.domain);
+    this.form.get('domain').setValue(param.domain);
 
     if (param.ipv4 === true) delete param.ipv4;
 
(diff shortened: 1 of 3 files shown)


src/app/components/form/form.component.ts
```diff
@@ -233,10 +233,18 @@ export class FormComponent implements OnInit, OnChanges, OnDestroy {
...
+  // Remove trailing spaces and dots, and leading spaces
+  private sanitizeDomain(domain: string): string {
+    return domain.replace(/[ \.]+$/, '').trim();
+  }
+
   public runDomainCheck(submitValid = true) {
     this.form.markAllAsTouched();
     let param = this.form.value;
 
+    param.domain = this.sanitizeDomain(param.domain);
```
pnax, line 245: The root `.` will be sanitized. Maybe prevent running the sanitize function if testing the root domain.

blacksponge: Oh yes, that's right

blacksponge: Should be better now

pnax (approved):

tgreenx (dismissed):

src/app/components/form/form.component.ts
```diff
@@ -233,9 +233,22 @@ export class FormComponent implements OnInit, OnChanges, OnDestroy {
     }
   }
 
+  // Remove trailing spaces and dots, and leading spaces
+  private sanitizeDomain(domain: string): string {
+    domain = domain.trim();
+    if (domain == '.') {
+      return domain;
+    } else {
+      return domain.replace(/[ \.]+$/, '');
```
mattias-p, line 242: I'm slightly surprised by this regex. It seems it would normalize `zonemaster.net... . ..  .` to `zonemaster.net`. Is this intentional? I would have expected to see something like `/\.? +$/`, but maybe that's overly strict?

mattias-p (approved):
Still this looks good to me.

src/app/components/form/form.component.ts
blacksponge: Yes it was intentional but I realise now that this may not be the best behaviour. I have updated it.

Also now I the domains are not updated in the form after being sanitized, this way inputting `zonemaster.net. . . `  then clicking submit twice will always result in the same `zonemaster.net. . ` domain and not `zonemaster.net. . ` then `zonemaster.net. `.

src/app/components/form/form.component.ts
```diff
@@ -239,7 +239,7 @@ export class FormComponent implements OnInit, OnChanges, OnDestroy {
     if (domain == '.') {
       return domain;
     } else {
-      return domain.replace(/[ \.]+$/, '');
+      return domain.replace(/\.$/, '');
```
mattias-p, line 242: The PR description says that trailing spaces should be stripped, but it looks to me that's not being done anymore.

blacksponge: It is actually performed before, as it is should also be done for the root.

https://github.com/zonemaster/zonemaster-gui/blob/5a2cd7e4c998ca5981c84bd80797cd640f142b57/src/app/components/form/form.component.ts#L238

mattias-p (approved):
LGTM

pnax (approved):

Issue #279, opened 2021-11-25
Do not rely on external CSS when exporting results to HTML


pnax:
Currently we rely on an external bootstrap CSS file when exporting the results to HTML.
It would be nice to have the few style rules hard coded within the HTML exported file.

https://github.com/zonemaster/zonemaster-gui/blob/8062b2f779fcdea655c056fc324359dfb3788642/src/app/components/result/result.component.ts#L196

matsduf:
Do we really need that many different export formats? I think we should focus on one or two and skip the rest. zonemaster-cli also, indirectly, offers export formats.

pnax:
HTML is already a supported export format.

matsduf:
But do we need it? Is it used in any noticable volume? All formats will need some maintenance. If we remove 1-2 formats that means less to maintain.

pnax:
> Is it used in any noticable volume?

I think we won't have any answer to this question.

pnax:
Closed with #308

PR #308, opened 2022-04-12
avoid using bootstrap for html export

blacksponge:
## Purpose

Remove dependency on external bootstrap in HTML export.

## Context

#279 

## Changes

Uses inline style instead of bootstrap.

## How to test this PR

Export a test result in HTML and open the export. The look should be close enough to the one using bootstrap.

### DIFF
--- a/src/app/components/result/result.component.ts
+++ b/src/app/components/result/result.component.ts
@@ -128,7 +128,7 @@ export class ResultComponent implements OnInit, OnDestroy {
       // TODO clean
 
       this.test = {
-        id: data['id'],
+        id: data['hash_id'],
         creation_time: new Date(data['creation_time'] + 'Z'),
         location: document.location.origin + this.location.prepareExternalUrl(`/result/${domainCheckId}`)
       };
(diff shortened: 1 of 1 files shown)


src/app/components/result/result.component.ts
```diff
@@ -227,20 +227,48 @@ export class ResultComponent implements OnInit, OnDestroy {
 
         const result = `
           <!doctype html>
-          <html class="no-js" lang="${this.language}">
+          <html lang="${this.language}">
             <head>
               <meta charset="UTF-8">
-              <!--[if IE]><meta http-equiv="X-UA-Compatible" content="IE=edge"><![endif]-->
               <meta name="viewport" content="width=device-width, initial-scale=1.0, shrink-to-fit=no">
               <title>Zonemaster TEST</title>
```
matsduf, line 234: I suggest that the title is changed to "Zonemaster Test Result" and if possible and simple also add the domain name in the title, e.g. "Zonemaster Test Result iis.se"

matsduf (commented):
I have not run this yet, and I will do that. Else it is probably fine.

pnax (dismissed):
Tested and it's working nicely.

matsduf (approved):
I approve.

PR #319, opened 2022-05-23
use created_at instead of deprecated creation_time in result

blacksponge:
## Purpose

* `creation_time` is deprecated, use `created_at` instead
* avoid workaround when parsing dates

## Context

* Use new API property from https://github.com/zonemaster/zonemaster-backend/pull/967

## Changes

Use `created_at` instead of `creation_time` 

## How to test this PR
* run a test
* check that the time of the test is correctly displayed and in the local timezone.

### DIFF
--- a/src/app/components/result/result.component.ts
+++ b/src/app/components/result/result.component.ts
@@ -131,7 +131,7 @@ export class ResultComponent implements OnInit, OnDestroy {
 
       this.test = {
         id: data['hash_id'],
-        creation_time: new Date(data['creation_time'] + 'Z'),
+        creation_time: new Date(data['created_at']),
         location: document.location.origin + this.location.prepareExternalUrl(`/result/${domainCheckId}`)
       };
 


mattias-p:
Have you considered pushing this until next release? That way the new version of GUI could be compatible with the old version of Backend. (Unless the new GUI is already relying on new features in Backend, of course.) This would give our users greater flexibility in the presence of bugs, and I think that's nice.

mattias-p:
I just like having flexibility. In some places you can anticipate that flexibility is going to be needed. But it's not uncommon that flexibility is needed or desired in a place where you couldn't or didn't anticipate it. When flexibility is easy to achieve we shouldn't throw it away.

If we're reasonably sure that https://github.com/zonemaster/zonemaster-gui/pull/316/files#diff-7e0b06dd0b9f241872f9c86aa9948803b9c8e2dc94f0914a6ef6c7f85c7a2ffcR39 is the only place where we rely on a new feature from Backend, I think it would make sense to revert that line for this release and reintroduce it for the next one. But if we don't know and it's difficult to find out what those changes or to revert them, it might not be worth the effort.

pnax (approved):

matsduf (approved):

Issue #309, opened 2022-04-12
Make it possible to find old tests without running a new test


matsduf:
If you want to inspect an old test you have to run a new test first. This is a limitation of the GUI, not the Backend. Make it possible to provide a domain name and then look for already run tests.

PR #333, opened 2022-07-18
Make it possible to look for a domain history without running a test

pnax:
## Purpose

This PR is about providing an easy way to fetch previous results for a given domain.
A new page has been created that provides a form similar to the domain check's form to look for a test's history. The history list is displayed on the page reusing the modal (from the result page).

Two new routes are created:
* `/history` that displays the page with the history form
* `/history/<domain>` that populates the form and queries for the history of \<domain\>.

zmgui-history2
zmgui-history3

## Context

Addresses #309 

## Changes

* New route `/history`
* New route `/history/<domain>`
* New "History" page with a form
* The `sanitizeDomain()` method from the form component has been moved to an external file to make it available to the project (expand this method to convert the domain to lower case)

## How to test this PR

Open the History page. Here is a list of things that can be tested:
* get the history for a \<domain\> using the form
* access the history from the URL directly `/history/<domain>`
* open a result from the history page
* switch language from the history page
* check that the modal window is not altered (from the result page)
* play with the browser history

---

### 1. Display a secondary button
<details><summary>Screenshots</summary>

image
image

</details>


pnax:
Of course, since the GUI has evolved, the tests are breaking. This makes me think, I haven't created any tests for this feature. I'll provide some.

matsduf:
Will "Check" and "History", respectively, really be understandable? What about "Run a new test" and "View completed tests"?

pnax:
> Will "Check" and "History", respectively, really be understandable?

I think both are fine as a start.

> What about "Run a new test" and "View completed tests"?

I feel like there are too many words, and Zonemaster does not always run a new test.

A screenshot with your proposal:

image

matsduf:
> > Will "Check" and "History", respectively, really be understandable?
> 
> I think both are fine as a start.

Well, I disagree. Especially "history" I find problematic.

> > What about "Run a new test" and "View completed tests"?
> 
> I feel like there are too many words, and Zonemaster does not always run a new test.

From the screen shot I do not think the number of words is a problem. Technically, it could sometimes be the latest test. So "Check" and "View completed tests" then?

pnax:
> Especially "history" I find problematic.
> From the screen shot I do not think the number of words is a problem.

I think we are trying to address something with a very subjective point of view. During F2F we decided to do some deep work on the GUI in 2023/2024. Hopefully this would be done with the help from UX/UI experts. In the meantime I find it hard to come up with the best wording.

matsduf:
> I think we are trying to address something with a very subjective point of view. During F2F we decided to do some deep work on the GUI in 2023/2024. Hopefully this would be done with the help from UX/UI experts. In the meantime I find it hard to come up with the best wording.

Now the interface is quite simple. There is one field to fill in and one button to press. The exact wording on that button is less important because there is no choice. With this change there will be two buttons to choose between, and suddenly it is important what the buttons say, and maybe the color, to make sure that the interface does not get too complex.

What do you think of adding a radio button under "Options" under "o Disable IPv4"? 

E.g.
```
o List already completed tests, do not run a new tesst
```

matsduf:
I really think that moving the feature done to the options section is the correct approach. The feature will not be used very much, and therefore it should not increase the complexity of the interface for the common usage.

matsduf (changes requested):
I really think that moving the feature done to the options section is the correct approach. The feature will not be used very much, and therefore it should not increase the complexity of the interface for the common usage.

blacksponge:
The radio button is a bad approach in my opinion, viewing the test history is not an parameter for the test but a completely different action. I do think though that the wording should be improved, I propose `Check domain`, `View history`. The color for the history button should not be orange in my opinion, it is a secondary action and should have more discreet color (light grey or something). Also I feel that the vertical padding is a bit too much. It was like that to fit the input height but is not a necessity anymore, but I don't have strong feeling about that so it is up to you.

On a side note the option form is only used for testing a domain, I think the wording there should also be changed, e.g. `Test options` or maybe better `Test parameters`.

matsduf:
> The radio button is a bad approach in my opinion, viewing the test history is not an parameter for the test but a completely different action. 

I think you have a point. First I reacted on the wording, then I felt that the alternative of viewing test history came too much in focus.

Today, you fill the domain in and press the only button there is. With the proposed change, you have to decide which of the two buttons to press, which is increased complexity for the newcomer or seldom-comer. That is why I suggested a radiobutton (or could be a check box) in the options  section.

> I do think though that the wording should be improved, I propose `Check domain`, `View history`. The color for the history button should not be orange in my opinion, it is a secondary action and should have more discreet color (light grey or something).

What about keeping the check button as is, and adding a check box under the domain field with some text saying that a new check should not be run, only give a list of completed tests.

> On a side note the option form is only used for testing a domain, I think the wording there should also be changed, e.g. `Test options` or maybe better `Test parameters`.

I agree that "Test options" would be better than today's "options".

blacksponge:
> What about keeping the check button as is, and adding a check box under the domain field with some text saying that a new check should not be run, only give a list of completed tests.

I am not sure about that. If we want to keep this form straightforward and simple what about adding a page "History" accessible from the navigation menu that is just about looking up hold tests for a domain name? Typically I see this to be like the proposed "History" page screenshot but with a input and button bellow the title page. Like this we would also avoid having to go back to look up another domain and we avoid confusion between the two feature.

pnax:
I moved the feature to another page "History" accessible from the navigation menu.

matsduf:
In general, this could work, but I find the terms "Domain check" and "History", respectively, in the top menu problematic. Both are refer to domain check. Could "New check" and "Completed checks" work better?

When you press "Domain check" you get:
image
It would make more sence if "Domain name" is replaced by **Run a domain check**.

When you press "History" you get:
image
Why is the button blue instead of red? I think it should be the same color in both cases.

I would make more sense if "History" is replaced by **List completed checks** and if the button says **List checks**.

Below "Domain check" the following text is shown, but not below "History". I think it should be the same text in both cases.
image

There is room for more text. Above "About Zonemaster" there could be a short paragraph describing what will happen in the two cases. In the case of normal check it could make sense to mention undelegated test and "options".

pnax:
> Why is the button blue instead of red?

I recall that when discussing about this PR, there was a suggestion to use a different color for the History button. But I might have missed something.

> There is room for more text. Above "About Zonemaster" there could be a short paragraph describing what will happen in the two cases. In the case of normal check it could make sense to mention undelegated test and "options".

This is a good idea, I've open #338 for that, so that this PR focuses on this new history feature.

matsduf:
> > Why is the button blue instead of red?
> 
> I recall that when discussing about this PR, there was a suggestion to use a different color for the History button. But I might have missed something.

When both were in the same tab it might have made sense, but when they are in different tabs it is better to have a uniform look and feel.

> > There is room for more text. Above "About Zonemaster" there could be a short paragraph describing what will happen in the two cases. In the case of normal check it could make sense to mention undelegated test and "options".
> 
> This is a good idea, I've open #338 for that, so that this PR focuses on this new history feature.

That is reasonable.

Issue #200, opened 2021-04-23
Better integration with other services


mattias-p:
I learned today that Hardenize links to three other tools from its [result pages](https://www.hardenize.com/report/zonemaster.net/1619171893#domain_dnssec). Among those tools is zonemaster.net :tada:

All three tools provide different levels of integration:
* Zonemaster.net - It's just a link to an empty domain name field. You have to enter the domain name you ran Hardenize on and press enter or click the start button.
* DNSSEC Analyser - It's a link to a pre-filled domain name field. You have to focus the field and press enter. (Weirdly enough they don't seem to have a start button.)
* DNSViz - If there is an older test result you're immediately redirected to the result. If there is no test result you're presented with a pre-filled domain name field and an analyze button.

We should make it possible to create links with the domain name pre-filled, and perhaps even automatically start the test when the domain name is pre-filled. We should also make this feature discoverable by vendors. Perhaps by adding a "Integration" documentation topic that can be easily found by someone looking for it.

Some care should be taken when designing the URL schema for this feature. In particular the schema should allow for extension with other options.

E.g. I believe it would make sense to extend this feature to also allow the inclusion of fake delegation details in the link. The idea here is to give DNS providers an easy way to let their end-users test their NS-record and DS-record updates before making the jump and hoping for the best.

matsduf:
This is a good idea. The first step is the design and specification of the URL schema. It should be documented as an API to make sure that we do not change anything without an explicit decision.

PR #334, opened 2022-07-19
Perform a test through the URL directly

pnax:
## Purpose

It has been discussed to be able to easily pass a domain to test to Zonemaster GUI, without interacting with the form. The idea is to pass the domain name in the URL `/check/<domain>`.
The form's input field will be populated with this domain name and the form will be sent.

No option are currently passed. Only the domain. As I see it, we could extend this work to include options as GET parameters.

## Context

Addresses #200 and follow-up on #335

## Changes

* New route `/check/<domain>`
* Redirect `/domain_check` to `/check`
* Redirect `/test/<test-id>` to `/result/<test-id>`

## How to test this PR

* access `/check/<domain>` and make sure the test is started for \<domain\>
* once on the result page, go back to the previous page, the form should be sent again and the results page should be loaded
* e2e tests have been added to test both redirections

### DIFF
--- a/src/app/components/form/form.component.ts
+++ b/src/app/components/form/form.component.ts
@@ -7,6 +7,7 @@ import {
   Validators,
   AbstractControl} from '@angular/forms';
 import { Title } from '@angular/platform-browser';
+import { ActivatedRoute, Params, Router } from '@angular/router';
 import { LangChangeEvent, TranslateService } from '@ngx-translate/core';
 import { Subscription } from 'rxjs';
 
@@ -51,14 +52,27 @@ export class FormComponent implements OnInit, OnChanges, OnDestroy {
   private _showProgressBar: boolean;
   private langChangeSubscription: Subscription;
 
+  private domainName: string;
+  private routeParamsSubscription: Subscription;
+
   public history = {};
   public test = {};
   public disable_check_button = false;
   public form: FormGroup;
 
-  constructor(private formBuilder: FormBuilder, private translateService: TranslateService, private titleService: Title) {}
+  constructor(private activatedRoute: ActivatedRoute,
+    private formBuilder: FormBuilder,
+    private translateService: TranslateService,
+    private titleService: Title) {}
 
   ngOnInit() {
+    this.routeParamsSubscription = this.activatedRoute.params.subscribe((params: Params) => {
+      if ( params['domain'] ) {
+        this.domainName = params['domain'];
+        console.log(this.domainName);
+      }
+    });
+
     this.langChangeSubscription = this.translateService.onLangChange.subscribe((_: LangChangeEvent) => {
       if (this.form.touched) {
         this.runDomainCheck(false);
@@ -76,6 +90,9 @@ export class FormComponent implements OnInit, OnChanges, OnDestroy {
 
   ngOnDestroy() {
     this.langChangeSubscription.unsubscribe();
+    if ( this.routeParamsSubscription ) {
+      this.routeParamsSubscription.unsubscribe();
+    }
   }
 
   private static atLeastOneProtocolValidator(control: AbstractControl) {
--- a/src/app/components/form/form.component.ts
+++ b/src/app/components/form/form.component.ts
@@ -52,7 +52,6 @@ export class FormComponent implements OnInit, OnChanges, OnDestroy {
   private _showProgressBar: boolean;
   private langChangeSubscription: Subscription;
 
-  private domainName: string = '';
   private routeParamsSubscription: Subscription;
 
   public history = {};
@@ -71,8 +70,8 @@ export class FormComponent implements OnInit, OnChanges, OnDestroy {
 
     this.routeParamsSubscription = this.activatedRoute.params.subscribe((params: Params) => {
       if ( params['domain'] ) {
-        this.domainName = params['domain'];
-        this.form.controls.domain.setValue(this.domainName);
+        let domainName: string = params['domain'];
+        this.form.controls.domain.setValue(domainName);
         this.runDomainCheck();
       }
     });
--- a/src/app/components/form/form.component.ts
+++ b/src/app/components/form/form.component.ts
@@ -69,7 +69,6 @@ export class FormComponent implements OnInit, OnChanges, OnDestroy {
     this.routeParamsSubscription = this.activatedRoute.params.subscribe((params: Params) => {
       if ( params['domain'] ) {
         this.domainName = params['domain'];
-        console.log(this.domainName);
       }
     });
 
@@ -78,8 +77,13 @@ export class FormComponent implements OnInit, OnChanges, OnDestroy {
         this.runDomainCheck(false);
       }
     });
+
     this.generate_form();
     this.titleService.setTitle('Zonemaster');
+
+    if ( this.domainName ) {
+        this.runDomainCheck();
+    }
   }
 
   ngOnChanges(changes: { [property: string]: SimpleChange }) {
--- a/src/app/components/form/form.component.ts
+++ b/src/app/components/form/form.component.ts
@@ -66,9 +66,14 @@ export class FormComponent implements OnInit, OnChanges, OnDestroy {
     private titleService: Title) {}
 
   ngOnInit() {
+    this.titleService.setTitle('Zonemaster');
+    this.generate_form();
+
     this.routeParamsSubscription = this.activatedRoute.params.subscribe((params: Params) => {
       if ( params['domain'] ) {
         this.domainName = params['domain'];
+        this.form.controls.domain.setValue(this.domainName);
+        this.runDomainCheck();
       }
     });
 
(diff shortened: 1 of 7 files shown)


matsduf:
Would it be relevant to be able to include the language setting? /domain_check/lang=<lang>/<domain> or /domain_check/<lang>/<domain>. The former is more flexible and could be opened for eg. /ns=<ns> in the future

matsduf:
I also want to come back to comment in the issue (https://github.com/zonemaster/zonemaster-gui/issues/200#issuecomment-825571535): "This is a good idea. The first step is the design and specification of the URL schema. It should be documented as an API to make sure that we do not change anything without an explicit decision."

pnax:
> Would it be relevant to be able to include the language setting?

I don't think it is. Since the language is only used to display the results, it does not impact how a test is run.

> The first step is the design and specification of the URL schema.

This PR proposes a simple design for the URL schema that can be easily extended. Is this good enough to document this feature?

matsduf:
> > Would it be relevant to be able to include the language setting?
> 
> I don't think it is. Since the language is only used to display the results, it does not impact how a test is run.

That is true, but the result is then displayed, and that should be in some language. If you create a link from the .fr registry to test your domain it would be reasonable to show the result by default in French.

We have actually already got a request for making it possible to include language setting in the URL to Zonemaster. 

At least we should make sure we do not close the doors for other options such as language and settings for undelegated test. And that different options might be combined.

> > The first step is the design and specification of the URL schema.
> 
> This PR proposes a simple design for the URL schema that can be easily extended. Is this good enough to document this feature?

We already have /result/<hash-id> being used. That should be included in the documentation. Do we have anything else?

docs/API.md
```diff
@@ -0,0 +1,15 @@
+# API
+
+The Zonemaster GUI provides a small API to access specific resource and perform
+specific tasks.
+
+The supported paths are the following:
+
+* `domain_check`: the default path, access the input form
+* `domain_check/<domain>`: populate the input form with `<domain>` and launch
```
matsduf, line 9: Or shorter just `check`?

I can see that we want to add other parts to the `check`. We do not have to do it now, but I suggest that is we make room for it now. One possibility is to define that from first "/" to the end or next "/" there is one or more keys.

Keys are strings of 'a-z' and can take an value, if defined for the key. The value is a string of 'a-z0-9'. The value is separated from the key by '='. 

Keys are separated by ':'. 

Unknown keys can be ignored.

Now we only define 'check', 'result', 'test' and 'faq'. In the future we might add a key 'lang' that can be combined with 'check', e.g. `/check:lang=en/afnic.fr` or `/faq:lang=fr` or `/result:lang=fi/<test-id>`

pnax: > Or shorter just check?

I kept the original path.

Why not keeping it simple and use query strings?
```
domain_check?lang=en
faq?lang=en
domain_check/afnic.fr?lang=fr&no-ipv4&ns=...&ds=...
```

> We do not have to do it now

I agree. This is left to a later work.

pnax: > Why do you think "domain_check" is better than just "check"?

I didn't say it's better. I'm just saying that this is already the default path. So why not just reusing it for a start?

> That could be used for the API in general, couldn't it?

I think so yes.

> Is there any advantage of /domain_check/afnic.fr instead of /?domain_check=afnic.fr?

The only advantage is to build on what already exists.

pnax: The goal here is not to define a whole API for the GUI. It is only to extend what we currently have to allow populating the input form from the URL and automatically launch the test. A side effect of this PR is to create a new API.md document.

matsduf: > I didn't say it's better. I'm just saying that this is already the default path. So why not just reusing it for a start?

I thought that this was about creating something new. How is "domain_check" the default? What would the consequence being of changing that to "check"?

> > Is there any advantage of /domain_check/afnic.fr instead of /?domain_check=afnic.fr?
> 
> The advantage I see is to build on what already exists.

Do you mean exist as "/result/" etc? But now we should have a broader perspective where we see that we in the future may want to add other elements, and then you suggest them to be added as "?KEY" or "?KEY=VALUE", which is reasonable. Why not base the new API on that? What would the consequence being of using that instead?

matsduf: > The goal here is not to define a whole API for the GUI. It is only to extend what we currently have to allow populating the input form from the URL and automatically launch the test. A side effect of this PR is to create a new API.md document.

I disagree that the API specification is a side effect. Instead of adding features ad-hoc we should specify first in a way that make the API stable for future extensions.

pnax: > I thought that this was about creating something new.

The goal of this PR is to provide an easy way to launch a test from the URL.

> I disagree that the API specification is a side effect. Instead of adding features ad-hoc we should specify first in a way that make the API stable for future extensions.

This is not the goal of this PR.

matsduf: We could break out the creation of the API document to a separate PR and pause this PR until the API has been set.

matsduf:
You have not given any good reason to why the path should be `domain_check/<domain>` instead of `check/<domain>` more than it is simpler because it is there. It is of course better to have a short path and we have to think about the future and the future use.

The fact that we today have both `result` and `test` as paths shows why it is important to document and discussed before something is implemented.

In the code it looks like `history` is also a path used by the GUI. Could you explain?

I can accept the proposal as a start if
1. `domain_check` is changed to `check` (or something else as short) in the API.
2. Just as with `test` and `result`, `domain_check` probably has to be kept as an alias (can be in bookmarks). In both cases they should preferably be rewritten to `result` and `check`, respectively.
3. the implementation will be done in such a way that it will be possible to extend it with modifiers, e.g. `?lang=en`, and also multipel modifiers, e.g. `?lang=en&ns=ns1.example.com`.

We have already had a request from a main user on being able to include language setting in the call.

pnax:
> We have already had a request from a main user on being able to include language setting in the call.

Could you point to this request? I can't find it.

> In the code it looks like history is also a path used by the GUI. Could you explain?

Nope. It's been here [since the beginning](https://github.com/zonemaster/zonemaster-gui/commit/143f0abeaea0d2f96f9274b720b7509271167824). Maybe there was an idea to make it possible to easily access a test history at that time already. Maybe something else. Who knows...

matsduf:
> > We have already had a request from a main user on being able to include language setting in the call.
> 
> Could you point to this request? I can't find it.

I will try to find it. It was probably in a mail directly to me.

> > In the code it looks like history is also a path used by the GUI. Could you explain?
> 
> Nope. It's been here [since the beginning](https://github.com/zonemaster/zonemaster-gui/commit/143f0abeaea0d2f96f9274b720b7509271167824). Maybe there was an idea to make it possible to easily access a test history at that time already. Maybe something else. Who knows...

OK. Thanks. It then seems like https://github.com/zonemaster/zonemaster-gui/blob/master/src/app/app.module.ts#L54 could be removed without any side effects. And I think it should.

matsduf:
> I've rebased and applied the suggestions from #335.

Do you think I should merge now or should we wait for more reviewers?

matsduf (approved):
This looks fine. Tested, and all works as it should. I have only tested with a default base path, not with a custom path.

#335 should be merged first.

src/app/components/form/form.component.ts
```diff
@@ -51,21 +52,38 @@ export class FormComponent implements OnInit, OnChanges, OnDestroy {
...
       if (this.form.touched) {
         this.runDomainCheck(false);
       }
     });
+
     this.generate_form();
     this.titleService.setTitle('Zonemaster');
+
+    if ( this.domainName ) {
+        this.runDomainCheck();
```
blacksponge, line 85: The `runDomainCheck` should be call from the subscribe callback, otherwise it will not be triggered if route parameter is changed (e.g. changing the domain in url and going back in history to the first one).

src/app/components/form/form.component.ts
```diff
@@ -76,6 +94,9 @@ export class FormComponent implements OnInit, OnChanges, OnDestroy {
 
   ngOnDestroy() {
     this.langChangeSubscription.unsubscribe();
+    if ( this.routeParamsSubscription ) {
```
blacksponge, line 97: Is there a case where this is not defined?

blacksponge (commented):
Nice feature (also thank you for removing the underscore in the url), I still have a couple of comments

src/app/components/form/form.component.ts
pnax: fixed

src/app/components/form/form.component.ts
pnax: Not that I'm aware of, I removed the test.

src/app/components/form/form.component.ts
```diff
@@ -51,21 +52,37 @@ export class FormComponent implements OnInit, OnChanges, OnDestroy {
...
+    private titleService: Title) {}
 
   ngOnInit() {
+    this.titleService.setTitle('Zonemaster');
+    this.generate_form();
+
+    this.routeParamsSubscription = this.activatedRoute.params.subscribe((params: Params) => {
+      if ( params['domain'] ) {
+        this.domainName = params['domain'];
+        this.form.controls.domain.setValue(this.domainName);
```
blacksponge, line 75: Is the `domainName` property used anywhere else? If not would it make send to remove it?

matsduf (approved):

matsduf (approved):
#335 has been merged and as far as I can see this could be merged.

PR #337, opened 2022-09-13
Use a drop down menu to select between languages

pnax:
## Purpose

Zonemaster supports more and more languages, and the GUI displays all of them on a row. This could lead to eating a lot of space. This PR replaces this row with a drop down menu (a \<select\> element) with an icon representing language selection (icon from [fork awesome](https://forkaweso.me/Fork-Awesome/icon/language/)).

image

## Context

Addresses a comment made when discussing the possibility to provide a new page for the "History" (PR #333) where it was pointed out that in some locale, the menu will span a lot of available space (almost eating on the language selection space).

## Changes

Remove the horizontal language list in the header.

## How to test this PR

Use the new drop down language selection menu to switch across languages. The page should be updated accordingly.

### DIFF
--- a/src/app/components/navigation/navigation.component.html
+++ b/src/app/components/navigation/navigation.component.html
@@ -22,6 +22,7 @@
       </ul>
 
       <div class="lang">
+        <span class="icon"><i class="fa fa-language" aria-hidden="true"></i></span>
         <div class="d-block">
           <select name="lang" [(ngModel)]="lang" (change)="setLanguage(lang)" class="form-select" id="languageSelection">
             <option *ngFor="let itemLang of enabledLanguages" [lang]="itemLang" [value]="itemLang" [selected]="lang === itemLang">
--- a/src/app/components/navigation/navigation.component.html
+++ b/src/app/components/navigation/navigation.component.html
@@ -21,8 +21,11 @@
         </li>
       </ul>
 
-      <div class="lang">
-        <span class="icon"><i class="fa fa-language" aria-hidden="true"></i></span>
+      <div class="lang" i18n-title title="Change language">
+        <label for="languageSelection" class="icon">
+            <i class="fa fa-language" aria-hidden="true"></i>
+            <span class="sr-only" i18n>Change language</span>
+        </label>
         <div class="d-block">
           <select name="lang" [(ngModel)]="lang" (change)="setLanguage(lang)" class="form-select" id="languageSelection">
             <option *ngFor="let itemLang of enabledLanguages" [lang]="itemLang" [value]="itemLang" [selected]="lang === itemLang">
--- a/src/app/components/navigation/navigation.component.html
+++ b/src/app/components/navigation/navigation.component.html
@@ -22,15 +22,8 @@
       </ul>
 
       <div class="lang">
-        <div class="d-none d-sm-block">
-          <ng-container  *ngFor="let itemLang of enabledLanguages; let i = index">
-            <a [lang]="itemLang" (click)="setLanguage(itemLang)" [title]="languages[itemLang]" [class.selected]="lang===itemLang">{{ itemLang | titlecase }}</a>
-            <ng-container *ngIf="i < enabledLanguages.length - 1; else endLangNav"> | </ng-container>
-          </ng-container>
-          <ng-template #endLangNav> &nbsp; </ng-template>
-        </div>
-        <div class="d-sm-none">
-          <select name="lang" [(ngModel)]="lang" (change)="setLanguage(lang)" class="form-control" id="languageSelection">
+        <div class="d-block">
+          <select name="lang" [(ngModel)]="lang" (change)="setLanguage(lang)" class="form-select" id="languageSelection">
             <option *ngFor="let itemLang of enabledLanguages" [lang]="itemLang" [value]="itemLang" [selected]="lang === itemLang">
               {{ languages[itemLang] }}
             </option>
--- a/e2e/FR05.e2e-spec.ts
+++ b/e2e/FR05.e2e-spec.ts
@@ -18,16 +18,16 @@ test.describe('Zonemaster test FR05 - [Supports internationalization]', () => {
   ];
 
   for (const { language, code, expected } of testSuite) {
-    test(`should have ${language} language button`, async ({ page }) => {
-      const langNavLink = page.locator(`.lang > div > a[lang="${code}"]`);
+    test(`should have ${language} language option`, async ({ page }) => {
+      const langNavLink = page.locator(`select#languageSelection > option[lang="${code}"]`);
       await expect(langNavLink).toHaveCount(1);
     })
 
-    test(`should switch switch to ${language}`, async ({ page }) => {
+    test(`should switch to ${language}`, async ({ page }) => {
       await setLang(page, code);
       await expect(page.locator('h1')).toHaveText(expected);
 
-      const langNavLink = page.locator(`.lang > div > a[lang="${code}"]`);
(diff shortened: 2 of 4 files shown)


matsduf:
What does then language icon add? Is the width of the language menu fixed or adjusted to the longest language name?

matsduf (changes requested):
I think it looks fine. Instead of cryptic language codes you get language names in the language itself. I think that is an improvement. For this to be really good, the width of the field must adjust to the longest language name (and that is maybe already so).

I do not think that the language icon adds anything, on the contrary it distracts. I suggest that is removed. Else I can approve.

src/app/components/navigation/navigation.component.html
```diff
@@ -22,14 +22,8 @@
...
       <div class="lang">
-        <div class="d-none d-sm-block">
-          <ng-container  *ngFor="let itemLang of enabledLanguages; let i = index">
-            <a [lang]="itemLang" (click)="setLanguage(itemLang)" [title]="languages[itemLang]" [class.selected]="lang===itemLang">{{ itemLang | titlecase }}</a>
-            <ng-container *ngIf="i < enabledLanguages.length - 1; else endLangNav"> | </ng-container>
-          </ng-container>
-          <ng-template #endLangNav> &nbsp; </ng-template>
-        </div>
-        <div class="d-sm-none">
+        <span class="icon"><i class="fa fa-language" aria-hidden="true"></i></span>
```
blacksponge, line 25: Could you put the icon in a label linked to the select and provide an alternative text (in a `span.sr-only` for example) and a `title`?

blacksponge (commented):
I am fine with the icon, see comment for accessibility concern though.

pnax:
> For this to be really good, the width of the field must adjust to the longest language name (and that is maybe already so).

This is already so, you can test applying the following diff:

```diff
diff --git a/src/environments/common.ts b/src/environments/common.ts
index 122ee52..66ce50a 100644
--- a/src/environments/common.ts
+++ b/src/environments/common.ts
@@ -12,9 +12,10 @@ export const common = {
     'fi': 'Suomi',
     'fr': 'Français',
     'nb': 'Norsk (bokmål)',
-    'sv': 'Svenska'
+    'sv': 'Svenska',
+    'aa': 'A very long language name to test field width'
   },
-  enabledLanguages: [ 'da', 'en', 'es', 'fi', 'fr', 'nb', 'sv' ],
+  enabledLanguages: [ 'aa', 'da', 'en', 'es', 'fi', 'fr', 'nb', 'sv' ],
   configUrl: 'assets/app.config.json',
   pollingInterval: 5 * 1000,
 }
```

> I do not think that the language icon adds anything, on the contrary it distracts.

Hard to settle on the icon usage, I think some people might find it distracting as other might find it helpful (such as myself).

mvw-afnic:
I’m fine with the icon. I don’t find it distracting at all and as long as concerns about accessibility are addressed, I’m fine with it.

src/app/components/navigation/navigation.component.html
pnax: Sure, I've updated the code to make the language selection accessible.

pnax:
> Nothing happens if you click on it. It is not obvious what it "means".

I've just updated the code so that it's linked to the select (as suggested above). And I also provided a title, that is shown on hovering it.

src/app/components/navigation/navigation.component.html
```diff
@@ -21,8 +21,11 @@
           </li>
         </ul>
 
-      <div class="lang">
-        <span class="icon"><i class="fa fa-language" aria-hidden="true"></i></span>
+      <div class="lang" title="{{'Change interface language'|translate}}">
```
matsduf, line 24: Isn't "Change language" enough? Or "Select language"?

`src/assets/i18n/en.json` (and all other language files) should be updated too, shouldn't it, with the new string to translate?

I will rebuild and test.

pnax: > Isn't "Change language" enough? Or "Select language"?

This is to explain the purpose of the field, I think it's better to be specific. Could you expand why you think "interface" is too much here? 

> `src/assets/i18n/en.json` (and all other language files) should be updated too, shouldn't it, with the new string to translate?

Yes. I'd like to get the proposal validated before going into the translation.

matsduf: > > Isn't "Change language" enough? Or "Select language"?
> 
> This is to explain the purpose of the field, I think it's better to be specific. Could you expand why you think "interface" is too much here?

Yes, I see that, but in this context "interface" is not needed. I will be quite obvious that "change language" will change the language the result is presented in. Shorter is better. Furthermore, "interface" is a quite technical term that can distract users.

matsduf:
> I think that with the growing number of language we should consider moving forward with this proposal.

I agree that a drop-down menu with the language name is an improvement. Today's language codes are cryptic.

> > I still do not see the purpose of the icon. If I click on it, nothing meaningful happens. When hoover over it, I do get a tool-tip, that is partially covered by my mouse pointer (not seen in the screen shot). The text in the tool-tip is very small.
> 
> The icon make sense as it provides a visual aid to indicate where to change the language when an other language is selected. We could also make the dropdown expend on click but I do not think that this should be blocking this PR.

If the icon is self-explaining then it would be better not to show the menu until you have clicked on the icon. The title that comes up when hoovering on it is "Change interface language". The word "interface" does not add anything and can be hard to understand. Just "Change language" is shorter and easier to understand.

matsduf:
> I think that with the growing number of language we should consider moving forward with this proposal.

I agree that a drop-down menu with the language name is an improvement. Today's language codes are cryptic.

> > I still do not see the purpose of the icon. If I click on it, nothing meaningful happens. When hoover over it, I do get a tool-tip, that is partially covered by my mouse pointer (not seen in the screen shot). The text in the tool-tip is very small.
> 
> The icon make sense as it provides a visual aid to indicate where to change the language when an other language is selected. We could also make the dropdown expend on click but I do not think that this should be blocking this PR.

If the icon is self-explaining then it would be better not to show the menu until you have clicked on the icon. The title that comes up when hoovering on it is "Change interface language". The word "interface" does not add anything and can be hard to understand. Just "Change language" is shorter and easier to understand.

blacksponge:
>  If the icon is self-explaining then it would be better not to show the menu until you have clicked on the icon.

I think showing the two of them is better. I see the icon as a label of the select here, indicating that this select is used for language.

> The word "interface" does not add anything and can be hard to understand. Just "Change language" is shorter and easier to understand.

I don't mind if we change the wording. I also don't feel like "interface" adds much but I am fine either way.

matsduf:
I will review this again, but I can see that the develop branch has changed since this was created. Can this be rebased on current develop?

pnax:
> Can this be rebased on current develop?

Done.

blacksponge (approved):
Tested using mouse and keyboard, it is working as expected.

pnax:
Rebased on top of develop to fix a conflict.

matsduf (approved):

matsduf (commented):
But Travis complains. I can re-approve when that is fixed.

tgreenx:
LGTM. Tested (Firefox - Windows) and it works as expected.

tgreenx (approved):
LGTM. Tested (Firefox - Windows) and it works as expected.

pnax:
rebased, please re-review

blacksponge (approved):

matsduf (approved):

tgreenx (approved):

Issue #340, opened 2022-10-12
Betterize display of results in GUI


tgreenx:
A number of improvements could be done in how we present the results of a Zonemaster test (https://zonemaster.fr/result/).

Here is a snippet of how it looks today:

![image](https://user-images.githubusercontent.com/96772376/195355196-34a8686d-9d8b-44af-ad04-65cf3f193432.png)

We have a list of expandable menus for each Module (Basic, Connectivity, Address, etc...). Within each list, we get a message comprised of three elements: the Module name, the log level, and the (translated) message ID.

Instead, we could have:

- A new menu that summarizes the outcome of a Test Case:
   - Instead of the Module name (e.g. BASIC), we could (should) have the Test Case identifier (e.g. BASIC00), with a link to the Test Case specification (in the form of an hyperlink and/or a tooltip).
   - Instead of the log level name directly, we could have some sort of visual indicator (i.e. icon; check box, cross, warning sign, etc...) indicating the general outcome of a Test Case. There would be one icon defined for each log level, using the color code used today, and the icon used to summarize the outcome of that Test Case would be the one from the worst level (e.g a Test Case with 5 messages, 4 being `INFO` and one being `WARNING` would only show the `WARNING` icon).
   - Instead of the (translated) message ID, we could have the Test Case description (e.g. "Domain name must be valid")

=> In that way, the user can immediately grasp the outcome of a Test Case within each Module, without having to "decode" each message ID one by one.

=> The current detailed message IDs would still be present, hidden by default, and similarly accessible by an expandable menu on a lower level.

- Within the expendable menu of each Module, we could have the count of each outcomes for each Test Case within that Module. Counts for log levels equal to 0 could be hidden by default to get a better view.

=> Similarly, this gives the user an immediate view of the outcomes in each Module.

I made a drawing to illustrate what I have just written above.

![zonemaster-gui-idea](https://user-images.githubusercontent.com/96772376/195381324-6c6ec59b-13f6-4038-bfa2-8367b37e88a3.jpg)

Ideas for improvement welcome. A draft implementation will soon follow.

matsduf:
I like the idea. I also like that the test cases are more visible, and I think this could give some more structure to the output. And still keep the flexibility that we have in Zonemaster.

* The test case ID is already available in the JSON result blob for message.
* We probably have to rework the test case descriptions to make them more consistent and understandable, but that should not be too hard.
[...]

PR #341, opened 2022-10-12
New ui for test results

blacksponge:
## Purpose

Provide a better look and feel while browsing the results of a test.

## Context

#340 
https://github.com/zonemaster/zonemaster-backend/pull/1055
https://github.com/zonemaster/zonemaster-engine/pull/1144

## Changes

* Group messages by test case.
* Group test cases by modules.
* Changes severity level colors to improve contrast.
* Display test case id when test case is expanded.
* Filter buttons display a test case count instead of a message count.
* Test cases are sorted by severity, "UNSPECIFIED" have no header (always expanded) and sorted before "regular" test cases.
* Test case severity count badges.

Screenshot 2022-11-17 at 13-46-08 ppsfleet navy · Zonemaster

## How to test this PR

* Install the backend and engine from https://github.com/zonemaster/zonemaster-backend/pull/1055 and https://github.com/zonemaster/zonemaster-engine/pull/1144.
* Create a new test or check the results of an old one.


matsduf:
I disagree that the test case descriptions should live in GUI. It might be good for some POC, but not for a solution. The test case descriptions should be guaranteed to be the same version as the test cases executed, and that can only be achieved by putting the test case descriptions in Engine together with the messages.

matsduf:
Linking back to the test case specification requires some design. If linking is just done back to Github you would end up in latest version even if the installation was not latest version. Engine should probably signal, in the JSON blob, what version of Zonemaster that was run so that correct version of the specification can be pulled out. The question is also where the specifications should live. At Github or locally?

matsduf:
Another feature that we should consider is having the message tag available in the GUI, i.e. when doing mouse over the message in natural language, then the message tag would appear. In that way it would be easier to see which tag the message comes from and easier to find the tag in the specification.

matsduf:
@blacksponge, you ask for review, but is it possible to build it? Travis fails.

matsduf:
I ran a test build, but I cannot see any changes.

blacksponge:
I moved the test case description in engine (see https://github.com/zonemaster/zonemaster-backend/pull/1055 and https://github.com/zonemaster/zonemaster-engine/pull/1144 )

blacksponge:
> Error and critical are in black. I think that should be red.

Fixed.

> I think it would be better and less cluttered if the blue boxes with "(i)" and test case ID are not shown by default. Someone on the meeting suggested that the box only appears when the test case is expanded. I think that could be an alternative. Else the ID could come up when you hoover on the test case description.

I changed it so the test case id is shown only if the test case is expanded.

blacksponge:
I have changed a bit the design based on external inputs.

Because all test cases are display in the current implementation the result page is rather long. I am not sure I like having collapsing module sections as it hides what was actually tested and make test messages / test case id hidden in two level of collapsed blocks. I am thinking that maybe a summary before having the actual details could be an idea. The summary would include the test case counts by level and would allow the user to jump to a specific module by clicking on the module name.

If you think that could be an acceptable solution then I will implement it.

blacksponge:
For reference this is what I mean by "summary"

Example 1:
Screenshot 2022-10-27 at 14-21-49 dyn bortzmeyer org · Zonemaster

Example 2:
Screenshot 2022-10-27 at 14-21-33 ppsfleet navy · Zonemaster

blacksponge:
> In the new iteration the modules are already expanded (and cannot be collapsed), which gives a much longer list at start, at the cost of overview.

> What do you say about collapsing all test cases under their modules in the same way?

I thought about it. As we have now two level of grouping having everything collapsed at first means it is harder to go at the lowest level. Also I think I like having modules expanded by default. It allows to get a quickly jump to error without having to expand anything. But I agree that this make a much longer list.

Honestly I am not sure what would be the best solution here, this is why I am experimenting a bit with the UI.

matsduf:
I also tested 2092109 "blacksponge/poc-summary" but that was not an improvement. I still suggest that the modules are collapsed by default to give a short page at start. There should also be an indication of the highest severity level in each module.

blacksponge:
From https://github.com/zonemaster/zonemaster-gui/pull/352#issuecomment-1293717538
> This is too striped, like a fence.

I can see that, in general I like the summary idea and from the feedback I had offline @marc-vanderwal  too. If we can come up with a better idea that would be nice.

That being said, in the mean time I will make the modules collapsed again (even if I am not a found of multilevel collapsing...) and display the count of test cases per severity as a badge (somewhat like the drawing in #340)

matsduf:
> That being said, in the mean time I will make the modules collapsed again (even if I am not fond of multilevel collapsing...) and display the count of test cases per severity as a badge (somewhat like the drawing in #340) so that we can maybe ship the new UI this release.

I will test again when you have done that.

pnax:
> That being said, in the mean time I will make the modules collapsed again (even if I am not fond of multilevel collapsing...)

I also do like the fact the module are not collapsed by default. However I can live with the collapsed suggestion.

blacksponge:
I have finalize the new UI for the results, this is now ready for review.

pnax (approved):
Great proposal and work!

tgreenx (dismissed):
Refreshing new look! Well done.

matsduf:
Can you resolve these conflicts and I will test and review.

blacksponge:
I am thinking of merging #349 before, so I have resolved the conflicts there. Given the complicity of the changes I'd rather not do that several time...

matsduf:
> I am thinking of merging #349 before, so I have resolved the conflicts there. Given the complicity of the changes I'd rather not do that several time...

That could be fine. I cannot build #349, however.

blacksponge:
> Can you resolve these conflicts? And I will test and review.

The conflict have been resolved and e2e tests fixed.

matsduf (dismissed):

tgreenx (commented):
Very nice! Just one comment (non-blocking), maybe it is too much of a hassle for now:

Can tooltips which give the Test Case ID be hyperlinks to the actual specification in Github? i.e. https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/Basic-TP/basic00.md for Basic00

matsduf:
> Very nice! Just one comment (non-blocking), maybe it is too much of a hassle for now:
> 
> Can tooltips which give the Test Case ID be hyperlinks to the actual specification in Github? i.e. https://github.com/zonemaster/zonemaster/blob/master/docs/specifications/tests/Basic-TP/basic00.md for Basic00

We have discussed that, and we agreed to postpone that for v2023.1 release. It is not enough to link to the specification, it should be the correct version of the specification. And also, should those be stored on Github, or on the server serving the GUI?

pnax (dismissed):

tgreenx (dismissed):

blacksponge:
I have merged develop in this branch, please re-approve.

tgreenx (approved):

matsduf:
I will re-test.

matsduf (approved):

matsduf (approved):

PR #342, opened 2022-10-13
Better keyboard navigation

blacksponge:
## Purpose

Some elements on the "run test" page and "result" page were not focusable or if they were not actionable.
This PR tries to improve overall keyboard accessibility.

## Context

n/a

## Changes

Full keyboard navigation is now possible across all pages.
* Advanced option switches is now a styled checkbox and behave as one.
* Make the module "boxes" in the results focusable and expandable when Enter key is pressed.
* Log level filtering can be toggled using keyboard.
* The history button is now a actual button and not a link.
* Filter buttons in history modal are now button and not a broken radio input anymore.
* Some css cleanup.

## How to test this PR

Try to use the gui without a mouse.

### DIFF
--- a/src/app/components/form/form.component.css
+++ b/src/app/components/form/form.component.css
@@ -1,24 +1,8 @@
-.fa-play-circle-o:hover {
-  color: #fff;
-}
-
-.fa-play-circle-o {
-  color: #9ad;
-}
-
 #domain_check_name {
   font-size:16px;
   height:64px;
   color:#6F787E;
   border-radius:3px
-}
-
-.checkboxForm {
-  width: 100%;
-}
-
-.half-with {
-  width: 50%;
 }
 
 button.launch {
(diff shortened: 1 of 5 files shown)


matsduf:
I wanted to test, but I failed to build. See #343.

matsduf (changes requested):
1. I can run a test and select "history", but how do I select the history list that I want to view?
2. If I press Enter on "options" nothing happens.
3. I can select "filter text" but I cannot select the all-info-notice-warning-error-critcal buttons.
4. How do I select FAQ?

matsduf:
> I tested it and I did it like this:
> 
>     1. Tab until "History" is selected.
Yes.

>     2. Press "Enter" to open the modal.
Yes.

>     3. Tab until the link to the test you want to see is selected.

First "X" is selected. Tab to "All". Tab to "Delegated". Tab to "Undelegated". Pressing "Enter" on any of these will do as expected.

Pressing tab again, and I am out of the modal. I am out of the Zonemaster page. I will end up on the web browser menu. This is Firefox.

In Chrome I come to the first in the history list, and it seems to work as you expect.

> > If I press Enter on "options" nothing happens.
> 
> By default checkbox are toggled using "Space" (this is a browser behaviour).

OK, then it works in both Firefox and Chrome.
 
> > I can select "filter text" but I cannot select the all-info-notice-warning-error-critcal buttons.
> 
> When I press tab, the filer buttons are focused after the "Share" button. Then I can press Enter on any of them to filter. I you can not do that we should investigate.

Works in Chrome, but not in Firefox.

> > How do I select FAQ?
> 
> When you press Tab just after loading the page. The focus start with the Zonemaster logo then moves to the navigation menu links. If when you press Tab the focus is past the navigation menu, use Shift + Tab to go back until you are focusing "FAQ".

Works in Chrome, but not in Firefox.

matsduf (commented):
If the issues I find are issues with Firefox rather than the solution I am ready to approve.

blacksponge:
> If the issues I find are issues with Firefox rather than the solution I am ready to approve.

Can it be a "Firefox on Mac" issue, I am also using Firefox, but on Linux.

matsduf (approved):

mattias-p:
I've successfully release tested this for v2022.2.

PR #346, opened 2022-10-19
Update Zonemaster and DNS descriptive texts

tgreenx:
## Purpose

This PR proposes an update to the descriptive texts regarding Zonemaster ("About Zonemaster") and the DNS ("About domain name system, DNS").

## Context

Those two texts are found on the front page in Zonemaster GUI (https://zonemaster.net/domain_check). IMHO, especially due to their prominent position, they should be improved as much as possible and so I tried to provide an update in that regard while keeping most of the current meaning and phrasing.

## Changes

- src/assets/i18n/en.json

## How to test this PR

Simply go to the landing page of the website and texts should be updated.


src/assets/i18n/en.json
```diff
@@ -1,8 +1,8 @@
...
+    "ABOUT_ZONEMASTER_TEXT": "Give your domain a complete checkup! Zonemaster helps you assess how your domain is doing. Besides, you also get a better understanding of how the Domain Name System (DNS) works. When a domain name is sent to Zonemaster for testing, the program investigates the state of the domain (sometimes called 'zone') from top to bottom of the DNS tree. Specifically, this is usually done by examining DNS data from the root ('.') to the TLD (Top-Level Domain; e.g. '.se') for that domain, and then finally through the DNS servers that contain authoritative information about the specified domain (e.g., 'zonemaster.se'). Zonemaster is also able to perform many other tests, such as, for example, checking DNSSEC signatures, that different hosts can be accessed and that IP addresses are valid. This is all to make sure that your domain runs as smoothly as possible.",
```
pnax, line 3: suggestion: use `.net` TLD and `zonemaster.net` domain in the example

pnax, line 3: "such as, for example": heavy, maybe only keep "such as" and remove "for example"

src/assets/i18n/en.json
```diff
@@ -1,8 +1,8 @@
...
+    "ABOUT_DNS_TEXT": "Most people who use the Internet have never reflected on what actually happens when one types an address (i.e., URL) in their web browser and then clicks the 'Enter' button. In short, the Domain Name System (DNS) is a database that associates domain names to Internet Protocol (IP) addresses, the same way a phone book associates names to phone numbers. An IP address is a unique series of numbers that identifies every computer that is connected to the Internet. It’s similar to the way every telephone has its own number in the telephone network. Such long series of numbers work fine for computers, but for people it is way more difficult to memorize. Thus, the DNS has been developed as an extra system to do this mapping transparently, for example in cases where we want to reach a web server on a specific site or an email server to send an email to.",
```
matsduf, line 2: > Most people who use the Internet have never reflected on what actually happens when one types an address (i.e., URL) in their web browser and then clicks the 'Enter' button. 

For this type of text "URL" does not add anything. On the contrary, it make it more complex.

> In short, the Domain Name System (DNS) is a database that associates domain names to Internet Protocol (IP) addresses, the same way a phone book associates names to phone numbers.

For this text "Internet Protocol" does not add anything. It is better to directly refer to IP address, better know by people.

> Thus, the DNS has been developed as an extra system to do this mapping transparently, for example in cases where we want to reach a web server on a specific site or an email server to send an email to.

Suggest:

"Thus, the DNS has been developed as an extra layer to map between names that we can remember and IP addresses used by the computers whenever we want to reach a web server on a specific site or use other services on Internet."

src/assets/i18n/en.json
matsduf: > (...) such as, for example (...)

Select one of them, e.g. "such as" or "for example". Her they mean the same thing.

matsduf (changes requested):

src/assets/i18n/en.json
tgreenx: Updated

tgreenx: Updated

src/assets/i18n/en.json
matsduf: I do not fully agree. More people have a picture of what "IP" is than "Internet Protocol". Actually the same is with "DNS" vs. "Domain Name System". It would be better with "About domain names and DNS".

I would be better to add "What do 'IP' and 'DNS' stand for?" as a new FAQ and there expand and explain.

tgreenx: > It would be better with "About domain names and DNS".

Agreed, for a title we should use the acronym. I will update on next review.

> More people have a picture of what "IP" is than "Internet Protocol". Actually the same is with "DNS" vs. "Domain Name System".

Agreed as well, acronyms are way more well-known.

Still, in a descriptive text such as this, I feel like giving the actual definition of an acronym (which immediately follows such definition, avoiding any confusion) is particularly appropriate and costless. Having an entry in the FAQ feels unnecessary to me. Let's hear what the others think.

pnax (approved):

src/assets/i18n/en.json
tgreenx: @matsduf If I were to inverse the position of the acronym with its definition, would that somehow read better for you? e.g `DNS (Domain Name System)` instead of `Domain Name System (DNS)`. Another option that will soon be possible would be to give the definition of the acronym in a mouseover (hover) element.

matsduf: Yes, `DNS (Domain Name System)` is better and easier to read when (as in this case) the abbreviation is better known than the spelled-out term.

matsduf: Mouse over could be another option, but that would make handling of the translations more complext, wouldn't it? In the translated text it might not be "Domain Name System". It might be "domännamnssystem".

tgreenx: Updated

src/assets/i18n/en.json
```diff
@@ -1,8 +1,8 @@
...
+    "ABOUT_DNS_TEXT": "Most people who use the Internet have never reflected on what actually happens when one types an address in their web browser and then clicks the 'Enter' button. In short, the DNS (Domain Name System) is a database that associates domain names to IP (Internet Protocol) addresses, the same way a phone book associates names to phone numbers. An IP address is a unique series of numbers that identifies every computer that is connected to the Internet. It’s similar to the way every telephone has its own number in the telephone network. Such long series of numbers work fine for computers, but for people it is way more difficult to memorize. Thus, the DNS has been developed as an extra layer to map between names that we can remember and IP addresses used by computers whenever we want to reach a web server on a specific site or use other services on the Internet (e.g., mail).",
```
matsduf, line 2: In a formal text like this it is recommended to use "it is" instead of "it's".

src/assets/i18n/en.json
```diff
@@ -1,8 +1,8 @@
...
+    "ABOUT_ZONEMASTER_TEXT": "Give your domain a complete checkup! Zonemaster helps you assess how your domain is doing. Besides, you also get a better understanding of how the DNS (Domain Name System) works. When a domain name is sent to Zonemaster for testing, the program investigates the state of the domain (sometimes called 'zone') from top to bottom of the DNS tree. Specifically, this is usually done by examining DNS data from the root ('.') to the TLD (Top-Level Domain; e.g. '.net') for that domain, and then finally through the DNS servers that contain authoritative information about the specified domain (e.g., 'zonemaster.net'). Zonemaster is also able to perform many other tests, such as checking DNSSEC signatures, or that different hosts can be accessed and that IP addresses are valid. This is all to make sure that your domain runs as smoothly as possible.",
```
matsduf, line 3: In this text both "domain" and "domain name" are used to mean the same thing. I think that "domain name" is the best option.

src/assets/i18n/en.json
```diff
@@ -1,8 +1,8 @@
...
     "About Zonemaster": "About Zonemaster",
-    "About domain name system, DNS": "About domain name system, DNS",
+    "About domain name system, DNS": "About DNS",
```
matsduf, line 5: I cannot see why key and value are differenth here. I suggest:
```
"About DNS": "About DNS",
```

matsduf (dismissed):
I accept as-is, but I have three suggestions to updates.

src/assets/i18n/en.json
tgreenx: Updated

src/assets/i18n/en.json
tgreenx: Updated

src/assets/i18n/en.json
tgreenx: Changing the key has other implications elsewhere in the code. Also that behavior will change when #349 is merged, so will not update for now.

matsduf (dismissed):

tgreenx:
Rebased on latest develop

tgreenx:
Note that this breaks all current (old) translations in locales files, they will need to be updated again for each language.

pnax (approved):

PR #349, opened 2022-10-24
Upgrade angular and bootstrap

blacksponge:
## Purpose

This PR upgrade to Angular 13 and Bootstrap 5.

## Context

* Angular 12 EOL in November 2022.
* ng-bootstrap ties the angular version to the bootstrap version. While we could have waited for Angular 14 to upgrade as there is one version of overlap to ease the migration I took the opportunity to upgrade it tool.

~~Note: I am marking this PR as draft because of #348~~ Migration of the i18n system has been included in the PR.

## Changes

* [x] Upgrade of angular and bootstrap and some minor dependencies.
* [x] Some dev dependencies have been removed as they are unused.
* [x] Migration to angular/localize (implies some changes on how language change is handle)
* [x] Remove all mention of ngx-translated.
* [x] Update e2e tests with new URL, add small dev server to serve all locales (angular dev server can only serve one local at a time)
* [x] Add documentation for translation
* [x] Update configuration (default language is now irrelevant in app config)
* [x] Update Apache example configuration
* [x] Update instruction to serve GUI from different base URL
* [x] Add instruction to use translation tools (poedit only)

During the migration to angular/localize, I tagged some strings that were not translated so that they could be translated.
Note: once this PR is merged we should create some issues to review the new translation files.

## How to test this PR

The GUI should behave as usual.

Known behaviour changes:
Changing the language reload the page causing:
* the domain form to be closed and reset if the user was editing it at the time of change
* the results modules to be collapsed if they were expanded

The Apache configuration has been updated to work correctly with the new release bundle architecture.

Without using Apache a simple server can be used to use the localized bundle by running `npm run e2e:test-server` (test configuration that uses zonemaster.net api).

Issue #348, opened 2022-10-24
Unmaintained dependency ngx-translate-extract causes error on message extraction with Angular 13+


blacksponge:
[ngx-translate-extract](https://github.com/biesbjerg/ngx-translate-extract) is not maintained anymore. There is an ongoing discussion about forking the project in the issues.

In the meantime this will break our translation process if we upgrade to Angular 13 and so it will block the upgrade ( #349) until resolved. (Angular 12 is EOL next month so I initially wanted to upgrade asap.)

Instead of waiting for a fork, an other option would be to migrate to [@angular/localize](https://angular.io/guide/i18n-overview) but this changes drastically the how translation is done.
* Instead of having the translation happening "live" in the browser, we build a bundle for each local. They will then be served under a separated base URL, e.g. `zonemaster.net/en`, `zonemaster.net/se`, etc.
* This design is (rightfully in my opinion) to make the final application "lighter" for the end-user as there is no need to load the translation library and everything.
* This also changes the format of the transation file from JSON to XML.
* Finally the language detection would be make at the web server level, using the `accept-language` header (see [example from Angular documentation](https://angular.io/guide/i18n-common-deploy#configure-a-server)). That being said I think the example given in the documentation is over simplistic as it only use the first language in the header and the syntax can be much more complex. I did not found any robust method to do language detection without using a actual server being the reverse proxy (be it Apache or Nginx).

I am open to feedback and suggestion on what our next step would be.

blacksponge:
@matsduf @mattias-p @pnax @tgreenx @mvw-afnic

matsduf:
> In the meantime this will break our translation process if we upgrade to Angular 13 and so it will block the upgrade ( #349) until resolved. (Angular 12 is EOL next month so I initially wanted to upgrade asap.)

Yes, that is important.

> Instead of waiting for a fork, an other option would be to migrate to [@angular/localize](https://angular.io/guide/i18n-overview) but this changes drastically the how translation is done.

Do you expect Angular/localize to be available and stable for at long time?

>     * Instead of having the translation happening "live" in the browser, we build a bundle for each local. They will then be served under a separated base URL, e.g. `zonemaster.net/en`, `zonemaster.net/se`, etc.

Having different base URLs (or language suffix) is not a problem as such, but it changes the API that we have set in [API.md] and when we have started to communicate that, to make it possible to URLs for Zonemaster with the domain to test, we do not want to change it again.

[...]

PR #350, opened 2022-10-25
Migrate to @angular/localize (POC)

blacksponge:
## Purpose

This PR is a PoC to demonstrate how @angular/localize would be integrated.

## Context

#348 
Based on #349 (from commit 8fb0e2d4a9a29004163e9113e60a824bc0adccf8)

## Changes

* Use @angular/localize to do localization
* Migrate French translation to new format as an example

TODO:

* Ability to change the language
* Remove remaining references to ngx-translate
* Remove references to JSON translation files
* Update default apache configuration
* Update various documentations

## How to test this PR

Having the dev server serving multiple language is not possible anymore, to serve a specific language run
```
npm run start ----configuration=<lang>
```
Only French is supported for now.


matsduf:
I compiled it an run it behind apache as usual. Switching between languages from the menu only affected the messages.

angular.json
```diff
@@ -7,10 +7,21 @@
       "root": "",
       "sourceRoot": "src",
       "projectType": "application",
+
+      "i18n": {
+        "sourceLocale": "en-US",
```
matsduf, line 12: Why "en-US"? It should rather be "en".

matsduf:
Is "src/locale/messages.fr.xlf" the equivalent of today's "src/assets/i18n/fr.json"? Then that will be a problem. It is much more complicated to manage.

blacksponge:
> I compiled it an run it behind apache as usual. Switching between languages from the menu only affected the messages.

I have not fixed the menu yet.

> Is "src/locale/messages.fr.xlf" the equivalent of today's "src/assets/i18n/fr.json"? Then that will be a problem. It is much more complicated to manage.

Yes it is. The translator needs to update / add the `target` tags with the corresponding string. These are file that are supposed to be managed using a dedicated editor. I was thinking that maybe it could be an idea to use a collaborative translation platform. This way it would be easier to track the missing translation message / do the translation and even integrate new languages without having translator having to deal with the project technicalities.

angular.json
blacksponge: Yes maybe it should, I took this from the documentation guide without thinking about it too much/

matsduf:
> > Is "src/locale/messages.fr.xlf" the equivalent of today's "src/assets/i18n/fr.json"? Then that will be a problem. It is much more complicated to manage.
> 
> Yes it is. The translator needs to update / add the `target` tags with the corresponding strings. These are files that are supposed to be managed using a dedicated editor. I was thinking that maybe it could be an idea to use a collaborative translation platform. This way it would be easier to track the missing translation message / do the translation and even integrate new languages without having translator having to deal with the project technicalities.

I am not sure what you mean by "collaborative translation platform". There should be one xlf file per language, shouldn't there? And the messages.xlf file is the source file. Each one should be edited separately, shouldn't they?

Both POeditor and poedit support the format. I have tried both, and it seems to work with poedit at least.

blacksponge:
> I am not sure what you mean by "collaborative translation platform".

If poeditor is https://poeditor.com then this is the kind of platform I meant. From what I see they have a free plan and integrate with github to fetch the translation file. This could maybe reduce the work for the translators.

> There should be one xlf file per language, shouldn't there? And the messages.xlf file is the source file.

Yes messages.xlf is the source file, then we need to create one xlf file per language. This can be done by copying messages.xlf to messages.<lang>.xlf.

> Each one should be edited separately, shouldn't they?

Yes.

matsduf:
> If poeditor is https://poeditor.com then this is the kind of platform I meant. From what I see they have a free plan and integrate with github to fetch the translation file. This could maybe reduce the work for the translators.

Yes, that is the POeditor I meant. I guess that is what is used today, at least by some translators. I use Emacs in PO-mode for the PO files.

I tested POeditor and poedit for the xlf files. With POeditor I did not get the English string to translate from. That worked with poedit.

I guess this could work as good as the JSON file, or maybe even better since there are dedicated tools for it.

Issue #374, opened 2022-12-13
Update in Configuration.md needed


matsduf:
Quoted from https://github.com/zonemaster/zonemaster-gui/pull/332#issuecomment-1348269036:

> However the [documentation file](https://github.com/zonemaster/zonemaster-gui/blob/develop/docs/Configuration.md) needs an update, as the configuration is now per language (e.g. `dist/en/assets`) instead of global (i.e. `dist/assets/`) (due to https://github.com/zonemaster/zonemaster-gui/pull/349). A fix for that may also be desired.

matsduf:
Resolved by #375

PR #375, opened 2022-12-13
Deduplicate the assets directory

blacksponge:
## Purpose

Since the migration to @angular/localize one bundle is generated per locale. This duplicates the assets.

While for the images it is sub-optimal but acceptable, for the configuration it is not. A user should not be required to update 6 times (or more in the future) the configuration, it is unpractical and error prone.

This PR changes the way the release zip is built to only use one assets folder. The asset folder is colocated with the locales folder to form the following architecture.

```
└── dist
    ├── assets
    │   ├── css
    │   ├── faqs
    │   ├── favicon
    │   └── images
    ├── da
    ├── en
    ├── es
    ├── fi
    ├── fr
    ├── nb
    └── sv
```

This way the assets folder is only present once. Also it means that our previous documentation still applies as the configuration path as not changed. Users will not have to move/copy the configuration file else where and will be able to reuse the existing one.

## Context

#374 

## Changes

* Use only one asset directory for all locales.
* Correct a typo in documentation

## How to test this PR

* Deploy the GUI using the release zip and apache configuration.
* Images (ZM logo) should be displayed and configuration should be accessible (try setting a message banner for instance).
* FAQ should be accessible.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -323,7 +323,7 @@ from wich the GUI is served.
 
 
 **NOTE:** Don't forget to apply the changes to the `index.html` and Web
-server configurartion after each updates as thoses files will be overwritten.
+server configuration after each updates as thoses files will be overwritten.
 
 
 ## Change default language
--- a/scripts/create_release.js
+++ b/scripts/create_release.js
@@ -19,6 +19,12 @@ archive.pipe(output);
 
 archive.file('zonemaster.conf-example', { name: 'zonemaster.conf-example' });
 archive.file('LICENSE', { name: 'LICENSE' });
-archive.directory('dist/', 'dist');
+
+const localizedBundles = fs.readdirSync('dist', {withFileTypes: true})
+  .filter(entry => entry.isDirectory())
+  .map(entry => entry.name);
+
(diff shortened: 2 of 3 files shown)


tgreenx (approved):
Works as expected, and also when using a [custom base url](https://github.com/zonemaster/zonemaster-gui/blob/develop/docs/Installation.md#serving-the-gui-and-api-from-a-custom-base-url).
Tested as part of 2022-2 release testing on Debian 11.

matsduf:
@blacksponge, I am not against it, but what risks do you see doing this very late in the release cycle?

blacksponge:
> I am not against it, but what risks do you see doing this very late in the release cycle?

I did the smallest changes that I could to reduce the risk, and it has already been tested on two OSs. I do not forsee much risks coming from that but I still prefer to take it than deferring to next release and have something half finished for this one.

pnax (approved):
I think this is fine to include it within this release.

matsduf:
@blacksponge, is issue #374 fully resolved by this PR?

blacksponge:
> is issue https://github.com/zonemaster/zonemaster-gui/issues/374 fully resolved by this PR?

Yes it is

Issue #231, opened 2021-06-17
Replace the Github octocat in the GUI


mattias-p:
Replace the Github octocat in the upper right corner of the GUI with a "contribute" menu item.
We discussed this on a conference call at least a year ago and this is how I remember it.

mattias-p:
Fixed by https://github.com/zonemaster/zonemaster-gui/pull/403 and https://github.com/zonemaster/zonemaster-gui/pull/411.

PR #383, opened 2022-12-26
Remove GitHub logo from header

pnax:
## Purpose

This removes the Github logo from the header. There still is a link in the footer.

## Context

Partially addresses #231.

## Changes

* Remove the Github logo and link from the navigation component.
* Update text link to source code in footer 
* Update footer rendering

## How to test this PR

The logo should not be displayed in the upper right corner on wide screen.


blacksponge (commented):
I don't mind that the Github logo is removed from the header. It is indeed quite invasive. Maybe the text of the link in the footer could be changed to something more generic like "Source code" (or "Source code on Github" if we want to keep the Gihtub mention) or equivalent?

matsduf:
If Github does not require that there is a link to and mentioning of Github I do think that we need to have that. I suggest "Source code and documentation" as the link text. Github as such is not interesting.

src/app/components/footer/footer.component.html
```diff
@@ -1,20 +1,16 @@
...
-            <a class="text-dark" href="https://github.com/zonemaster/zonemaster">Github</a>
-          </div>
-        </div>
+      <div class="contact">
+        <i class="fa fa-envelope-o" aria-hidden="true"></i>
+        <a class="text-dark" href="mailto:{{contactAddress}}">{{contactAddress}}</a>
+      </div>
+      <div class="source-code">
+        <i class="fa fa-github" aria-hidden="true"></i>
+        <a class="text-dark" href="https://github.com/zonemaster/zonemaster" i18n>Source code</a>
```
matsduf, line 13: I think there is room for "and documentation" too.

blacksponge:
Replaced by #403 and #411

PR #384, opened 2022-12-26
GUI customization + dark theme

pnax:
## Purpose

This is about providing an easier way to customize the colors in the GUI using CSS variables. This also provides a dark theme for the GUI.

## Context

n/a 

## Changes

* centralize colors definition using CSS variables in `app.component.css`
* provide colors for light and dark theme
* new switch theme icon 

## How to test this PR

...


src/app/components/navigation/navigation.component.html
```diff
@@ -7,6 +7,11 @@
       <span class="navbar-toggler-icon"></span>
     </button>
 
+    <label for="theme-switch" class="icon theme-switch" i18n-title title="Switch light/dark theme">
```
blacksponge, line 10: I have several concerns about the current theme switching implementation (I do not claim I have a solution for all, though):
* having the checkbox associated to this label living in another component breaks the principles of component (they should be independent of each other), I know it breaks the "checkbox" hack but maybe prefer using event to propagate state from child to parent ;
* I don't like the fact that the label contains both light *and* dark, maybe change dynamically the label or just use "Switch theme";
* I am questioning the semantic of the checkbox, it feels associated to form and data input, as opposed to buttons that are more generic to dynamically change the content / apparent of the page. Also the checkbox semantic is checked / unchecked, with the current label, we don't know what check or uncheck means.;
* (more personal) I think I would prefer if we showed both the light and dark theme and highlight in some way the one in use. In the current implementation I had trouble figuring out if the icon was corresponding to the current theme or to the other one I could switch to. 
* Why keep the theme switch out of the menu when the language drop down is in the menu?
* The tab order is somewhat not intuitive, it is the first element focusable even though it is after the language select.
* The cursor should be a pointer to show that the element is interactive.

src/app/components/navigation/navigation.component.css
```diff
@@ -34,6 +48,14 @@ nav {
   height: 100%;
 }
 
+label.icon.theme-switch {
+    order: 3;
+    text-align: center;
+}
+label.icon.theme-switch .fa.theme-switch::before {
+    content: var(--zm-theme-icon);
```
blacksponge, line 56: Maybe this is better done in javascript.

src/app/app.component.css
```diff
@@ -2,4 +2,169 @@
...
+    --zm-body-dark-bg: #2C3034;
+    --zm-body-dark-color: var(--bs-light);
+
+    --zm-link-light-color: var(--bs-link-color);
+    --zm-link-light-color-hover: var(--bs-link-hover-color);
+    --zm-link-dark-color: #5a9dff;
+    --zm-link-dark-color-hover: #8ea5ff;
+
+    --zm-form-btn-bg-1: #ff5a5f;
+    --zm-form-btn-bg-2: #ff3c3a;
```
blacksponge, line 41: Maybe `--zm-form-btn-bg` and `--zm-form-btn-bg-hover` (as you did for other) are more explicit.

src/app/components/navigation/navigation.component.html
```diff
@@ -7,6 +7,11 @@
       <span class="navbar-toggler-icon"></span>
```
blacksponge, line 7: The burger is almost invisible in dark theme.

src/app/components/navigation/navigation.component.html
```diff
@@ -7,6 +7,11 @@
...
     </button>
 
+    <label for="theme-switch" class="icon theme-switch" i18n-title title="Switch light/dark theme">
+        <i class="fa theme-switch" aria-hidden="true"></i>
+        <span class="sr-only" i18n>Switch light/dark theme</span>
+    </label>
+
     <a class="navbar-brand" routerLink="/">
       <img src="{{logoUrl}}" alt="Zonemaster" class="logo">
```
blacksponge, line 16: The ZM logo is un readable in dark theme.

blacksponge (commented):
Nice feature, and this is a pretty good start! I still do have comments :)

Comments I couldn't associated to changes:
* The inputs (text input, select, check boxes) are still in "light" theme when dark "theme" is enabled. You might want to check [Bootstrap builtin dark mode](https://getbootstrap.com/docs/5.3/customize/color-modes/#dark-mode) to style bootstrap related components.
* Please support the [prefers-color-scheme media query](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme)
* Theme preference should be saved and loaded from (localStorage)[https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage] in a similar way that it is done for the language.

src/app/components/navigation/navigation.component.html
```diff
@@ -1,4 +1,4 @@
-<nav class="navbar navbar-light bg-light fixed-top navbar-expand-lg" [ngClass]="{'shrink': isShrunk}" #navView>
+<nav class="navbar fixed-top navbar-expand-lg" [ngClass]="{'shrink': isShrunk}" #navView>
   <a href="https://github.com/zonemaster/zonemaster/" class="github-corner d-none d-lg-block d-xl-block" aria-label="View source on Github">
```
blacksponge, line 2: The Github thing should have inverted colors in dark theme (Octocat on black with a white background)

blacksponge: (Ignore comments if https://github.com/zonemaster/zonemaster-gui/pull/383 is merged)

pnax:
Thanks for the thorough review. I'll update the PR with your comments.

PR #407, opened 2023-05-05
Moves HTML anchor into header and creates valid HTML

matsduf:
## Purpose

HTML files are created from markdown files. The current markdown code puts the anchor in a paragraph before the headline the link should link to. The solution does not create valid HTML. This PR resolves the issue.

## Changes

All FAQ files in `docs/FAQ/gui-faq-*.md`, but no change of content.

## How to test this PR

Links in table of contents should go to the right question.

### DIFF
--- a/src/app/components/faq/faq.component.css
+++ b/src/app/components/faq/faq.component.css
@@ -1,4 +1,3 @@
-:host ::ng-deep a[name^=q] {
-  position: relative;
-  top: -90px;
+.faq {
+  overflow-anchor: none;
 }
--- a/src/app/components/faq/faq.component.css
+++ b/src/app/components/faq/faq.component.css
@@ -1,4 +1,4 @@
-:host ::ng-deep span[id^=q] {
+.faq span[id^=q] {
   position: relative;
   top: -90px;
 }
--- a/src/app/components/faq/faq.component.css
+++ b/src/app/components/faq/faq.component.css
@@ -1,3 +1,8 @@
+:host ::ng-deep span[id^=q] {
+  position: relative;
+  top: -90px;
+}
+
 .faq {
   overflow-anchor: none;
 }
--- a/src/app/components/faq/faq.component.ts
+++ b/src/app/components/faq/faq.component.ts
@@ -1,4 +1,4 @@
-import { Component, OnInit, AfterViewChecked, OnDestroy, Inject, LOCALE_ID } from '@angular/core';
+Aimport { Component, OnInit, AfterViewChecked, OnDestroy, Inject, LOCALE_ID } from '@angular/core';
 import {ActivatedRoute} from '@angular/router';
 import { HttpClient } from '@angular/common/http';
 import { Subscription } from 'rxjs';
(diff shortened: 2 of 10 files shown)


marc-vanderwal (dismissed):
Looks good to me.

blacksponge:
Please update the translation guide https://github.com/zonemaster/zonemaster-gui/blob/master/docs/TranslationGuide.md#gui-faq-langmd

And the style of the FAQ https://github.com/zonemaster/zonemaster-gui/blob/master/src/app/components/faq/faq.component.css#L1

blacksponge:
`faq.component.css` needs to be updated as well, specifically the `ngAfterViewChecked` method,

```js
  ngAfterViewChecked(): void {
    if (!this.fragment) {
      return
    }

    const faqEntry = document.getElementById(this.fragment);
    if (faqEntry !== null) {
      faqEntry.scrollIntoView();
      this.fragment = '';
    }
  }
```

While testing I noticed a strange behaviour when clicking on a link for the first time on Firefox, the browser does not  scroll down all the way to the FAQ entry but stops with a warning in the console saying `Scroll anchoring was disabled in a scroll container because of too many consecutive adjustments`. This seems to be related to (scroll anchoring)[https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-anchor/Guide_to_scroll_anchoring].  Adding the following into the component style fixes the issue.
```css
.faq {
  overflow-anchor: none;
}
```

matsduf:
@blacksponge  and @rhedman, as far as I understand, the old content of src/app/components/faq/faq.component.css is not needed anymore.

blacksponge:
> as far as I understand, the old content of src/app/components/faq/faq.component.css is not needed anymore.

It is, the selector should be updated to `span[id^=q] ` instead of `a[name^=q] `.
This rule is to offset the scroll to avoid having the entry title hidden by the navigation menu at the top.

src/app/components/faq/faq.component.ts
```diff
@@ -1,4 +1,4 @@
-import { Component, OnInit, AfterViewChecked, OnDestroy, Inject, LOCALE_ID } from '@angular/core';
+Aimport { Component, OnInit, AfterViewChecked, OnDestroy, Inject, LOCALE_ID } from '@angular/core';
```
blacksponge, line 1: Extra `A`.

matsduf:
> > as far as I understand, the old content of src/app/components/faq/faq.component.css is not needed anymore.
> 
> It is, the selector should be updated to `span[id^=q] ` instead of `a[name^=q] `. This rule is to offset the scroll to avoid having the entry title hidden by the navigation menu at the top.

I will put it back, updated.

src/app/components/faq/faq.component.ts
matsduf: Fixed in conflict resolution.

rhedman:
@blacksponge @matsduf 
Could we create a setting in the config that completely turns off all styles? Something like: `globalStyles: false`. I'm implementing a new design and it's a hassle to always have to write CSS-selectors that overrides the default CSS.

matsduf:
@blacksponge and @rhedman, please rereview.

rhedman (changes requested):
Change this selector: :host ::ng-deep span[id^=q]
Into this .faq span[id^=q]

src/app/components/faq/faq.component.css
```diff
@@ -1,3 +1,8 @@
+:host ::ng-deep span[id^=q] {
```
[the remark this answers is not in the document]
matsduf: I have updated the CSS file.

rhedman (approved):

blacksponge (approved):

pnax:
### v2023.1 release testing

Clicking on the links in the FAQ scroll to the right section. The section title is properly displayed at the top of the viewport, just below the header.

PR #408, opened 2023-05-08
Make room for optional footer logo

matsduf:
## Purpose

If GUI is customized and the main logotype is replaced then there can be a need for a logotype in the footer. This PR makes room for that.

~~Since #403 also updates the footer, this PR is built on top of #403. Commit e914c53 belongs ot #403, not this PR.~~

The optional logo is selected in `app.config.json`. By default there is no such logo.

## How to test this PR

Verify that the footer works without the optional footer. With the footer additional updates of CSS have to be done, which is outside the scope of this PR.


src/app/components/footer/footer.component.html
```diff
@@ -1,36 +1,47 @@
...
       </div>
+
     </div>
 
-    <div class="row">
-      <div class="col-md-12">
-        <p class="copyright text-muted small" i18n>Copyright © AFNIC and The Swedish Internet Foundation. All Rights Reserved</p>
-      </div>
+    <div class="col-md-12" *ngIf="footerLogo !== ''">
+      <img src="{{footerLogo}}">
```
blacksponge, line 33: There is no alt text for the image. Either provide one from the configuration (should probably be language specific with a default language) or assume it is only cosmetic (probably not) and put an empty alt.

Also I don't see any constraint on size, on any screen width. What is the recommended size for the logo?

blacksponge:
I have merged #403 so you can rebase on it.

src/app/components/footer/footer.component.html
```diff
@@ -1,36 +1,47 @@
...
         </div>
       </div>
+
     </div>
 
-    <div class="row">
-      <div class="col-md-12">
-        <p class="copyright text-muted small" i18n>Copyright © AFNIC and The Swedish Internet Foundation. All Rights Reserved</p>
-      </div>
+    <div class="col-md-12" *ngIf="footerLogo !== ''">
```
blacksponge, line 32: Is it the wanted behaviour to have the logo below the footer? Can you provide a example how this is used?

I don't mind  having a optional logo in the footer I just want to understand how it integrates with the current UI if someone else wants to use it.

matsduf: We will use CSS to place and size the logotype in a good locatioin. We will replace the Zonemaster logo in the top with a Internetstiftelse logo in the top and put an Afnic logo in the footer.

src/app/components/footer/footer.component.html
```diff
@@ -1,36 +1,47 @@
 <footer class="pb-0 pt-2 bg-light">
   <div class="container">
-    <div class="row justify-content-between">
-      <div class="col-md-4">
-          <span placement="top" [ngbTooltip]="tipContent" i18n tabindex="0">Program versions</span>
-      </div>
-      <div class="col-md-5 pull-right">
-        <div class="row">
-          <div class="col-md-6">
+    <div class="col-md-12">
```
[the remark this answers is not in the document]
rhedman: > Why is this a (bootstrap) column (of the size of the container)?

To make sure it takes up the entire width and thereby ends up on top of each other

src/app/components/footer/footer.component.html
matsduf: Alt parameter added. I do not think translation is required. I guess what makes most sense is to have the same string as in the logotype, and that is not translated.

src/app/components/footer/footer.component.html
blacksponge: > We will use CSS to place and size the logotype in a good locatioin.

In this case it may be good to have some default CSS rules to prevent oversize and resize the logo if needed on smaller screen included in the pull request so that someone that want to use the feature does not have to hack the GUI to make it work. What do you think?

Example when I used the (clearly oversized) Zonemaster logo:
Screenshot 2023-05-09 at 17-02-16 Zonemaster

rhedman: > > We will use CSS to place and size the logotype in a good locatioin.
> 
> In this case it may be good to have some default CSS rules to prevent oversize and resize the logo if needed on smaller screen included in the pull request so that someone that want to use the feature does not have to hack the GUI to make it work. What do you think?
> 
> Example when I used the (clearly oversized) Zonemaster logo: Screenshot 2023-05-09 at 17-02-16 Zonemaster

Yeah, that's probably a good idea. Let's add a class e.g `footer-logotype` to the image and som default styles like

```
.footer-logotype {
    width: 100%;
    max-width: 150px;
    height: auto;
}
```

src/app/components/footer/footer.component.html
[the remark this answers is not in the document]
blacksponge: Okay, if it is equivalent maybe it would be easier to understand if it was a class "row". As I understand it, it would be the same visually, if not I am fine with it how it is.

src/app/components/footer/footer.component.html
```diff
@@ -30,7 +30,7 @@
     </div>
 
     <div class="col-md-12" *ngIf="footerLogo !== ''">
-      <img src="{{footerLogo}}">
+      <img src="{{footerLogo}} alt="{{footerLogoAlt}}">
```
blacksponge, line 33: Missing closing double quote after `{{footerLogo}}`

src/app/components/footer/footer.component.html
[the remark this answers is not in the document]
rhedman: > Okay, if it is equivalent maybe it would be easier to understand if it was a class "row". As I understand it, it would be the same visually, if not I am fine with it how it is.

Yeah it would be visually the same, but for clarity wrapping them in `row` would be more consistent with how the rest of the code is structured so I think we can go head and wrap them.

src/app/components/footer/footer.component.html
blacksponge: Thanks.

I was wondering, I don't see the contact address in your screenshot, do you hide it in CSS? If that is the case, maybe we should make it so that if it is empty in the configuration it is not rendered in the template.

src/app/components/footer/footer.component.html
matsduf: Fixed!

src/app/components/footer/footer.component.html
matsduf: I have set class to the footer logotype.

src/app/components/footer/footer.component.html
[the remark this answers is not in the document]
matsduf: > > Why is this a (bootstrap) column (of the size of the container)?
> 
> To make sure it takes up the entire width and thereby ends up on separate "rows". We could of course also wrap them in actual row-containers. But it's not a necessity. @matsduf We should change the classes to `col-12`. The `md` part is redundant.

Updated.

src/app/components/footer/footer.component.css
```diff
@@ -35,3 +35,10 @@ footer nav ul {
 footer nav li {
   list-style: none;
 }
+
+.footer-logotype {
```
blacksponge, line 39: Can you add a small margin bottom, so the logo is separated from the bottom of the screen?
e.g. `margin-bottom: 0.5rem;`

matsduf:
@blacksponge and @rhedman, please rereview.

pnax:
### v2023.1 release testing

Works as expected:
* by default, the footer does not display any other image
* used the `"footerLogo"` configuration to define the path to an image that is properly displayed another image
* used the `"footerLogoAlt"` item to define an alt text that is handled by the browser

PR #410, opened 2023-05-09
migrate to the new rpcapi methods

blacksponge:
## Purpose

Migrate to new RPCAPI methods.

## Context

Zonemaster Backend introduced new RPCAPI methods to replace the current
ones that are being deprecated. Most changes are renaming the methods
and encapsulating results in objects.

## Changes

Change of methods name
* version_info -> system_versions
* profile_names -> conf_profiles
* start_domain_test -> job_create
* test_progress -> job_status
* get_test_results -> job_results
* get_test_history -> domain_history
* get_data_from_parent_zone -> lookup_delegation_data

Methods that return a object but were not before:
* conf_profiles -> { profiles }
* job_create -> { job_id }
* job_status -> { progress }
* domain_history -> { history }

## How to test this PR

1. Set a custom public profile in backend
2. Check that the GUI displays the correct list of profiles
3. Query for delegation data, NS and DS, using the GUI advanced options
4. Create a new test, check that the progress bar is correctly updated, and result correctly fetched
5. Query for tests history

### DIFF
--- a/src/app/components/result/result.component.ts
+++ b/src/app/components/result/result.component.ts
@@ -245,8 +245,8 @@ export class ResultComponent implements OnInit, OnDestroy {
     if (!this.history) {
       this.alertService.info($localize `History information request is in progress.`);
 
-      this.dnsCheckService.getTestHistory(this.historyQuery).then(data => {
-        this.history = data as any[];
+      this.dnsCheckService.getTestHistory(this.historyQuery).then(({ history }) => {
+        this.history = history as any[];
         if (this.history.length === 0) {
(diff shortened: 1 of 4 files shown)


pnax:
### v2023.1 release testing

* custom profiles work as expected
* querying delegation data works with ASCII domain, when using unicode domain, no data is fetch. This is a bug already present in master
* same for DS data retrieval
* creating a test works as expected, the progress bar is updated and the results are displayed
* the history works as expected

matsduf:
This PR has been completely reverted by #428.

PR #434, opened 2023-08-22
Attempt to improve gui design

blacksponge:
## Purpose

Try to improve the overall look and feel and accessibility of the GUI.

## Changes

Highlights :

* Make the "about" lines shorter by making this area double columns and adding an decorative picture
* Improve color contrasts and make color usage more uniform
* Group inputs in fieldset and add visible label for all fields in advanced form
* Change FAQ layout: This change needed to move the content of the FAQ into the code and integrate it to the "Angular translation system" instead of having separate markdown files. This means that translated FAQ will need to be imported, knowing that I have changed some wording at the same time. ~~Also I took this oportunity to test crowdin.com integration, <https://crowdin.com/project/zonemaster-test>.~~
* Slightly bigger font size and line height
* Make document outline more consistent.
* Replace links that looks like buttons in result filter area
* Add labels and group fields in result filter area, add expand + collapse buttons

<details>
<summary>My todo lists</summary>

Things I want to do (in this PR)
- [x] Make "about" lines shorter
- [x] Improve contrast between active and not active navigation links
- [x] Make color usage more uniform
  - [x] disabled primary button color
  - [ ] secondary button color -- change or keep ?
  - [x] lighten danger and secondary button on hover instead of darkening them
- [x] Try to improve "about" information visualization in "about" text (use markup)
  - [x] About zonemaster
  - [x] ~~About DNS~~ need to change the text, not doing this now
- [x] Improve accessibility of advanced option form:
  - [x] make labels visible
  - [x] use a fieldset to group NS and DS inputs
- [x] Improve accessibility of result filtering form
  - [x] maybe replace links that look like buttons with checkboxes and labels in fieldset
  - [x] Move help / severity level info out of tooltips
- [x] Improve document outline
  - [x] replace h4/h5 with h2/h3 in about section
  - [x] Add section + headings for form  / result in result page
- [x] Improve "program version" (at least make the "program version" thing a link so it is an actionable item)
- [x] ~~Expand (+ highlight?) text in search result~~ not in this PR
- [x] Change FAQ layout
  - [ ] add "expand" button for each section -- surprisingly hard to do, need more thoughts

Things that I broke and need to fix
- [x] fix advanced form NS and DS inputs wrapping for medium / small screens
- [x] main heading for result page
- [x] fix message banner
- [x] Click on faq entry update url fragment
- [x] change of url fragment open faq entry
- [x] fix link in info alert in advanced form
- [x] i18n (french translation done)
[...]

PR #439, opened 2023-09-07
Add gui server

blacksponge:
## Purpose

Add a basic web server to serve GUI files.

## Context
Based on #434 
Following a discussion in the face to face, we decided to serve the GUI from a server to allow more flexibility.

## Changes

* [x] Add a server in perl (using Dancer2)
* [x] Template index.html to include additional resources and override default ones (for css)
* [x] Detect browser language using the server
* [ ] Add Makefile.PL for packaging
* [ ] Update reverse proxy documentation
* [ ] Update installation doc in zonemaster/zonemaster repo
* [ ] Add service script to manage daemon
* [ ] Add example configuration and update configuration documentation
* [ ] Remove unused files (configuration, scripts, ...)

## How to test this PR

Running the server:
```sh
% plackup scripts/zonemaster-gui
```


blacksponge:
This is still a draft as it is not finished, but I would like some early feedback to avoid going the wrong path.

blacksponge:
Commit [80e36dd](https://github.com/zonemaster/zonemaster-gui/pull/439/commits/80e36ddfa896af1ef280767ccabbb6df5e98bf28) is the relevant stuff

blacksponge:
I still need architecture approval on this PR to finish (or change) this work.

lib/Zonemaster/GUI.pm
```diff
@@ -0,0 +1,136 @@
...
+# Generate configuration
+get '/:lang/app.config.json' => sub {
+    send_as JSON => config->{client_config}
+};
+
+# Serve additional assets
+get '/:lang/additional/**' => sub {
+    my ( $path ) = splat;
+
+    if ( grep /^\.{2,}$/, @{$path} ) {
```
marc-vanderwal, line 26: This might not be sufficient. What happens if I request `/fr/additional/bogus/../../../../../etc/passwd` or something like that?

marc-vanderwal (commented):
In terms of architecture, I see no issues; I agree with the choice of Dancer2.

lib/Zonemaster/GUI.pm
blacksponge: When using `**` splat return an array of path component, one component per element, so if I am not missing anything, in your example `$path` would be `( 'bogus', '..', '..', '..', '..', '..', 'etc', 'passwd')` and the grep would work.

[Relevant documentation as reference](https://metacpan.org/dist/Dancer2/view/lib/Dancer2/Manual.pod#Wildcard-Matching)

marc-vanderwal: Oh okay, my bad.

matsduf:
Danser will create daemon listening to http, won't it? Will that be on localhost using apache as a reverse proxy? Will everything still be packaged in one zip file as today?

marc-vanderwal:
Dancer can spawn a small HTTP server for testing purposes, but when deploying a Dancer app in production, it can also spawn a FastCGI server, which is less expensive.

blacksponge:
> Danser will create daemon listening to http, won't it? Will that be on localhost using apache as a reverse proxy?

yes

> Will everything still be packaged in one zip file as today?

No it will be package as a perl module.

blacksponge:
I am delaying that for next release, I think this PR needs more thoughts.

PR #448, opened 2023-12-18
Avoid bundling the mock interceptor in prod env

hannaeko:
## Purpose

Exclude `mock.interceptor.ts` from the build file.

## Context

Since Angular 9, the trick used to exclude `mock.interceptor.ts` no longer works. This PR aims at correcting that.

## Changes

Import `mock.interceptor.ts` in the environment file instead of in `app.module.ts` .

## How to test this PR

Before the change, after running `npm run build` 
* The build size (as reported by the `npm run build` command) is `940.51 kB`
* Some test domains can be found in the resulting bundle
   ```
   % grep -ic progress.afnic.fr dist/en/main.*.js
   1
   ```
after,
* The build size (as reported by the `npm run build` command) is `914.67 kB`
* No test domain can be found in the resulting bundle
   ```
   % grep -ic progress.afnic.fr dist/en/main.*.js
   0
   ```


matsduf:
Can this be postponed to v2024.1 instead of adding it this late in the cycle?

marc-vanderwal (approved):
Looks fine to me.

PR #451, opened 2024-01-15
Update Norwegian translation (FAQ mostly)

hannaeko:
## Purpose

Fix the missing translation of the FAQ.

## Context

The translation system of the FAQ has changed. This import old translation into the new system.

## Changes

Norwegian translation.

## How to test this PR

**Note:** It felt to me that some text differs from the original in content for some FAQ items, maybe it should be reviewed and updated accordingly to be more inline with the other languages.


blacksponge:
@matsduf I don't know who does the Norwegian translation usually, could you add them as reviewer here please?

blacksponge:

PR #453, opened 2024-01-16
Update Spanish translation (FAQ mostly)

hannaeko:
## Purpose

Try to fix the missing translations of the FAQ.

## Context

The translations system of the FAQ has changed. This import old translations into the new system.

## Changes

Spanish translations.

## How to test this PR


blacksponge:

PR #454, opened 2024-01-16
Update Danish translation (FAQ mostly)

hannaeko:
## Purpose

Try to fix the missing translations of the FAQ.

## Context

The translations system of the FAQ has changed. This import old translations into the new system.

## Changes

Danish translations.

## How to test this PR

**Note**: There are some missing translation for the strings that had changed since the last FAQ translation. Also some may have drifted a bit so feel free to update where the translation could be improved to be more in line with the original text.


blacksponge:

PR #466, opened 2024-03-04
Cleanup old FAQ architecture files

hannaeko:
## Purpose

Remove unused FAQ files and scripts

## Context

Following #434 some files are not used any more as the FAQ now use the common translation system.

## Changes

* Remove `docs/FAQ/*`
* Remove `scripts/faq.js`
* Remove unused dependency `showdown` (markdown to html renderer)
* Update the translation guide

## How to test this PR

Building the release (`npm run release`) should work.


marc-vanderwal (approved):

mattias-p (approved):
I had a look around and I found src/assets/faqs/.gitkeep but no mentions of faqs being assets. Do we still need this directory or would it make sense to remove that file too?

marc-vanderwal:
> I had a look around and I found src/assets/faqs/.gitkeep but no mentions of faqs being assets. Do we still need this directory or would it make sense to remove that file too?

The build system still generates files in the `src/assets/faqs` directory, so this directory is still needed.

hannaeko:
Not anymore, the script that used it has been remove with this PR, I removed the .gitkeep, thanks for noticing :)

mattias-p (approved):

PR #482, opened 2024-09-25
Translation update

bzwitt:
## Purpose

Add slovenian translation

## Context

(e.g. Fixes #9999, Follow-up to #9999, etc.)

## Changes

...

## How to test this PR

...

PR #488, opened 2025-03-03
Preparation for v2024.2.1 release (GUI)

matsduf:
## Purpose

This PR is a preparation for v2024.2.1 release. 

## Changes
*  package.json
* src/environments/version.ts
* Changes

## Note

If #482 is updated (made complete) before this is merged, then the text in changes on #482 `Changes` could be updated.

## How to test this PR

Review.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,5 +1,17 @@
 Release history for Zonemaster component Zonemaster-GUI
 
+
+v4.4.0 2025-03-04 (part of Zonemaster v2024.2.1 release)
+
+ [Release information]
+ - Translations have not been fully updated in this release. They will
+   be updated in an upcoming release.
+
+ [Features]
+- Adds translation to Slovenian language. The translation is not
+  complete (#482)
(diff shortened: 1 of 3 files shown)


Changes
```diff
@@ -1,5 +1,17 @@
 Release history for Zonemaster component Zonemaster-GUI
 
+
+v4.4.0 2025-03-04 (part of Zonemaster v2024.2.1 release)
```
mattias-p, line 4: Maybe "introduced in" would be a better wording than "part of"? I'm thinking sometimes a component version is reused for multiple product releases.

mattias-p (approved):
Once #482 is merged I'm good with this.

Changes
matsduf: It is the same wording as last release. Let us look at the wording for next release. I think that "part of" is better than "introduced in", but maybe we could find a another wording that works better.

PR #521, opened 2025-12-11
Replace paraglide

tbleckert:
## Purpose

Replace the current outdated i18n solution Paraglide with a custom built Vite plugin. Reduces dependencies and easy to maintain.

## Context

Paraglide released a new major version that dropped support for SSG in Astro. While possible to make it work with some digging and manual work, we can see this as an opportunity to reduce some dependencies in favour of a simpler custom built solution. Paraglide was also indirectly affected by the Shai-Hulud worm as it had posthog as a dependency. 

## Changes

Introduces a custom built Vite plugin instead of Paraglide. Translation files stays intact except for the removal of the schema definition. Some references has been replaced as well.

Also upgraded Astro and other dependencies to remove all remaining vulnerabilities and have all packages up to date.

## How to test this PR

Just some general testing to make sure translations still work.

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -2,8 +2,16 @@ import { defineMiddleware } from 'astro:middleware';
 import { setLocale, isValidLocale } from '@/messages';
 
 export const onRequest = defineMiddleware(({ url }, next) => {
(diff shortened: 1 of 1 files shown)


marc-vanderwal (approved):
Looks good. Thanks!

PR #523, opened 2025-12-12
Rerender tests correctly

### LINKED ISSUE #522: New GUI : Test result messages do not load correctly when fetched from the history
_This relates to the new GUI implementation (#500)_
When selecting a previous test from the history, the URL and test date change correctly, but the test messages themselves do not. Refreshing the page loads the messages correctly.

tbleckert:
## Purpose

Fixes #522 

## Context

When navigating between test results, for example via the History dialog, the view didn’t update with the correct data. The issue was caused by a Svelte component relying on local state, which prevented it from reacting when the incoming data object changed.

## Changes

ResultInfo now uses a derived rune instead of local state, which ensures it updates whenever the passed data changes. The history navigation has also been updated to include the language prefix.

## How to test this PR

Run a test, then switch to an older one via the History dialog. The displayed data should update immediately and correctly. Also verify that reloading the page shows the same data.

### DIFF
--- a/src/lib/components/DomainTest/ResultInfo.svelte
+++ b/src/lib/components/DomainTest/ResultInfo.svelte
@@ -43,7 +43,7 @@
         query: '',
     });
     const rawData = $derived(data.results);
-    const result = $state(groupResult(data.results));
+    const result = $derived(groupResult(data.results));
     let showExport = $state(false);
     let showShare = $state(false);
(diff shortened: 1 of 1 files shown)


tgreenx:
@tbleckert thanks for the fix. could you fix the conflicts by rebasing this PR?

PR #542, opened 2026-04-09
All-In-One Docker Image

MichaelTimbert:
## Purpose

This pull request provides a simplified, all-in-one Docker image for users who want to run zonemaster-gui from a single command line.

This is not a production-ready image; it uses SQLite as its database and offers no configuration options.

Also see documentation in https://github.com/zonemaster/zonemaster/pull/1487

## Context

Related to https://github.com/zonemaster/zonemaster/issues/1283

## Changes

To be consistent with other zonemaster repositories, I created a Makefile to generate the docker image.

## How to test this PR

You must be able to build a Docker image with `make docker-build` (you need zonemaster/backend docker image for this).

`docker run --rm -d --name zm -p 8080:80 zonemaster/gui:local gui`

Go to `http://localhost:8080`, and Perform some tests and check if everything is working (history, FAQ, language, etc.).

Finally, you can stop the image with:
`docker stop zm`


docker/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+USER root
+COPY ./zonemaster_web_gui_${version}.zip .
+
+RUN apk add apache2 apache2-proxy
+
+RUN install -vd /var/www/html/zonemaster-web-gui
+RUN install -vd /var/log/zonemaster
+RUN unzip -d /var/www/html/zonemaster-web-gui zonemaster_web_gui_${version}.zip
+RUN rm -f zonemaster_web_gui_${version}.zip
```
marc-vanderwal: Each individual RUN command will create an intermediate layer, so it’s better to combine these four RUN commands into one, with `&&` stringing together each individual operation, like so:


```suggestion
RUN install -vd /var/www/html/zonemaster-web-gui /var/log/zonemaster && \
    unzip -d /var/www/html/zonemaster-web-gui zonemaster_web_gui_${version}.zip && \
    rm -f zonemaster_web_gui_${version}.zip
```

matsduf:
Looks good to me, but is Docker/Dockerfile sensitive to trailing space after backslash?

matsduf:
Please rebase on updated develop.

matsduf:
@MichaelTimbert Can you create a PR that adds documentation in both "docs/public/using" and in https://github.com/zonemaster/zonemaster/blob/develop/docs/internal/maintenance/ReleaseProcess-create-docker-image.md?

matsduf:
I want to test to create an image on one server and running it on another.

Makefile
```diff
@@ -1,11 +1,11 @@
 VERSION = $(shell jq -r '.version' package.json)
 
 docker-build:
-	docker build --tag zonemaster/gui:local --build-arg version=$(VERSION) -f docker/Dockerfile .
+	docker build --tag zonemaster/aio:local --build-arg version=$(VERSION) -f docker/Dockerfile .
```
matsduf: I feel that aio is too cryptical. Will that really mean anything to the users? "gui" is maybe not good, but then "all-in-one"? Is there a problem with a name that long?

marc-vanderwal: I agree. When I see, “aio”, the first thing that comes to my mind is “asynchronous input/output”.

Nextcloud has a Docker image named [nextcloud/all-in-one](https://hub.docker.com/r/nextcloud/all-in-one), so that should be fine.

MichaelTimbert: I agree with you, "all-in-one" is a better name.

marc-vanderwal (approved):

matsduf (approved):
I have succeeded to make it run on Linux, och there I created an image file that I took to a Macbook, but there I could start but not connect. I cannot even see that it is listening to the port.

I guess this is good enough for now. Has it been tested on Windows?

matsduf:
I can confirm that zonemaster-all-in-one runs on Docker (Docker Desktop) on (updated) Windows 10. There is still no support for IPv6 when using Docker Desktop. As far as I understand a limitation sits in WSL (Windows sub-system for Linux).

tgreenx:
> I can confirm that zonemaster-all-in-one runs on Docker (Docker Desktop) on (updated) Windows 10. There is still no support for IPv6 when using Docker Desktop. As far as I understand a limitation sits in WSL (Windows sub-system for Linux).

Apparently there is, you need to change a configuration file though: see official documentation https://docs.docker.com/engine/daemon/ipv6/#use-ipv6-for-the-default-bridge-network and also e.g. https://github.com/nextcloud/all-in-one/blob/main/docker-ipv6-support.md#docker-desktop-windows-and-macos

